PR #28, opened 2021-01-14
Updated SIOP request encrypted response params to identify encryption key

jdkizer9:


dleve123:
I collaborated a bit with @jdkizer9 on this.

To ensure context is shared on GH, the problem here:

As the spec currently stands, it is ambiguous which of the Issuer's encryption keys a Holder should use to encrypt the SIOP Response sent back to the Issuer.

Might as well just specify the encryption KID for the key the Issuer wants the data to be encrypted with in the SIOP Request. 

This is very similar to the data exchange that happens when the [Holder requests Health Cards via FHIR from the Issuer](https://healthwallet.cards/#lab-results-are-finalized), note the inclusion of the `encryptForKeyId` parameter.

madaster97:
I don't think this change should be made, since the `id_token_encrypted_response_kid` field doesn't exist in the [OIDC Dynamic Registration Metadata spec](https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata). That's what the 
request's `registration` field is based on.

The correct behavior for this should be:

1. SIOP Request Object species the algorithm in `id_token_encrypted_response_alg`
2. OP resolves the RP's DID Document, and arbitrarily picks a key that matches the algorithm
3. In the JWE jose header, the OP would indicate what key it used via the `kid` field
    - It should be REQUIRED for the RP's DID Document to have a kid set for their public encryption key. 
    - Unclear how this is conveyed in the spec, but having no kid defined a DID document with several encryption keys would cause ambiguity.

Example JWE Header for encrypted SIOP response payload:
```js
{
  "alg": "ECDH-ES",
  "enc": "A256GCM"
  "kid": "did:ion:<<identifer for RP>>#<<encryption-key-id>>"
}
```

I think this conversation can be more generally applied to DID SIOP, and not specifically the health cards spec. That said, I think it's important we spell this out by addressing the JWE header TODO in this [section](https://healthwallet.cards/#connect-health-wallet-to-lab-account) of the health cards spec. I'm working on updates to address that.

madaster97:
Correction here. In the definition of the `add-public-keys` [Patch Action](https://identity.foundation/sidetree/spec/#add-public-keys), a public key entry is required to have an `id` element. 

**Example add-public-key patch**
```json
{
  "id": "signing-key-1",
  "purposes": [
    "authentication",
    "assertionMethod"
  ],
  "type": "JsonWebKey2020",
  "publicKeyJwk": {
    "kty": "EC",
    "alg": "ES256",
    "crv": "P-256",
    "x": "qRLha9zHWJotjOlf7tmzfdtg_CN3YPLXHH7Fir2tb3Y",
    "y": "eIGAdtuxLXl6VMfzyiSkv065HWnaoHsbYB45EqWlnoQ"
  }
}
```

This has a few implications:
1. In the Jose header of a JWT, we know the `kid` points to a specific `publicKeys[i].id` element if the `kid` is presented in DID format `did:ion:<<identifer for JWT signer>>#<<signing-key-id>>`. See point 3 for why `kid` SHOULD be a <DID>#<id> reference, even for signed SIOP request objects.
2. The actual JWK embedded patch's a `publicKeys[i].publicKeyJwk` element may have a `kid`, but it doesn't seem to have any direct use.
3. If a **SIOP RP** provides a `registration.jwks_uri` in their request, the `kid` of JWTs signed by OR encrypted for this requester MAY reference a specific `kid` from that URI in non-DID format. This seems to be covered in the DID SIOP spec, in the [SIOP request section](https://identity.foundation/did-siop/#generate-siop-request). 
   - **Disclaimer** I don't think we'd recommend this workflow, since SIOP openid providers have no way of trusting a `jwks_uri` in advance. It seems recommended NOT to provide a jwks_uri in SIOP request, and to have `kid`s reference DID documents in all cases.

docs/index.md
```diff
@@ -260,7 +260,7 @@ And a payload like:
   "state": "<<client-supplied value, possibly empty>>",
   "registration":  {
     "id_token_signed_response_alg" : "ES256",
-    "id_token_encrypted_response_alg": "ECDH-ES",
+    "id_token_encrypted_response_kid": "<<encryption-key-id>>",
```
jmandel, line 263: If we eliminate `id_token_encrypted_response_alg` we may wan to add a note that the `kid` must refer to a key with alg=`ECDH-ES`

madaster97: I don't think we should add this field. It would be breaching the OIDC spec (and the DID SIOP spec). There's likely been conversation about this in the Dynamic Client Registration spec before, and that seems like a better place to pitch this change.

jmandel:
@madaster97 great points re: availability of kids and potential for conflict with jwks_uri-based keys. I think we want to be prescriptive and mandate the `did:ion:<<identifer for JWT signer>>#<<signing-key-id>>` format for kid.

madaster97:
@jmandel just to be clear, which `kid` are you suggesting be a rigorous DID:
1. A new `id_token_encrypted_response_kid` field in the registration metadata (which only solves the problem for SIOP responses and not VCs)
2. The `kid`s in the JWE Jose headers of encrypted SIOP responses and VCs. Example:
```json
{
  "alg": "ECDH-ES",
  "enc": "A256GCM",
  "kid": "did:ion:<<identifer for encrypter>>#<<encryption-key-id>>"
}
```

I'm not sure if point 2 is already explicitly mandated, but I do think that is the best solution.

jmandel:
TL;DR: @jdkizer9 would you be okay if we closed this PR without merging?


---
Thanks for the clarification @madaster97. I meant (2) and I think we agree -- if it's not explicitly mandated, it should be. (Reviewing our current language, I think it's well covered, actually. So we're OK.)

If we *did* merge this PR, then we should also stipulate (1) -- but I'm not convinced we have a strong use case for it , so would prefer to leave it out. (To be fair, there are potential scenarios that the OIDC spec doesn't contemplate, like an org with a single DID but multiple hardware devices each managing its own key. Imagine, say, a guest check-in situation where a hotel controls 10 iPads where guests check in; if the org wanted to have one DID with 10 separate keys, then it'd be important for the SIOP request to name a key available on the device initiating the request. And OIDC didn't really design for this scenario. So... we have a theoretical use case, which could matter in the future, but which pushes boundaries and requires "inventing" things. As such, I'd rather leave it out of scope for now.)

jmandel:
P.S. @madaster97 would you mind sending me a DM on https://chat.fhir.org?

dleve123:
Hmm, I'm perhaps a bit confused by https://github.com/smart-on-fhir/health-cards/pull/28#issuecomment-761630754,

Just to make sure I'm understanding some of this terminology:

> OP resolves the RP's DID Document, and arbitrarily picks a key that matches the algorithm

- OP == Self-Issued OpenID Provider == SIOP == Health Wallet (in this context).
- Therefore, RP == Relying Party == Issuer == Lab (in this context)

- As such,

we can transform the above quote to: 

> Health wallet resolves the lab's DID Document, and arbitrarily picks a key that matches the algorithm


If so, then your proposal is:

1. Empower the Wallet to choose any public key **that matches the requested algorithm** resolved from the lab's DID document to encrypt the SIOP Response and specify that keys `id` within the `kid` of the JWE header. Is that correct?

2. Make no changes to the `$HealthWallet.issueVc` FHIR operation.

----

While I of course can't speak for @jdkizer9, we worked on this together and this seems good to me. I like the consistency of always having the JWT header specify the keys needed to decode/verify. This proposal does perhaps push more logic on the wallet to decide upon the public key, but doesn't seem too onerous.

Would be great if either of you could confirm my understanding!

jmandel:
Thanks for the detailed question @dleve123  -- and yes your understanding matches what I am proposing here. (Now that you have worked through this interpretation, if there are places we can update the language in our specification to make it easier for future readers to catch on more quickly, that kind of "fresh eyes" feedback is always welcome )

jdkizer9:
Sorry @jmandel, I was out Thursday and Friday so just getting caught up on this now. If I'm reading this right, wallets can use the `alg` field in the DID document public keys to discover _any_ encryption key to use to encrypt the SIOP response. If that's right, it sounds like a better and more standards compliant approach than what I proposed and I'm happy to close.

jmandel:
Thanks @jdkizer9. Will close, and if down the line we ever need sub-did-level encryption-key-hinting we can re-introduce this capability.

Issue #29, opened 2021-01-15
Typos & Clarifications


madaster97:
# Overview
Reading through the docs I've run into a couple places where we could be clearer.

# List
- [x] Health Wallet as [Issuer](https://healthwallet.cards/#install-a-health-wallet-app): In linked section, we say the wallet needs to generate a key *to enable verification of JWT signatures created by **this issuer***. Could we clarify that **issuer is referring to the holder signing JWS**, not an actual *Issuer* from the [conceptual model](https://healthwallet.cards/#conceptual-model)?
- [x] Clarify [SIOP Request Payload iss DID](https://healthwallet.cards/#connect-health-wallet-to-lab-account): In the example **DID SIOP Request**, the JOSE header has a `kid` of `did:ion:<<identifer for lab>>#<<verification-key-id>>`, and the payload has an `iss` claim of `did:ion:<<did-value>>`. Should the iss actually be `did:ion:<<identifer for lab>>` for clarity, since this JWT was issued by the lab?
- [x] Clarity the phrase *"as JWE or JWS"* ([example](https://healthwallet.cards/#lab-results-are-finalized) as VCs), in that this is decided by the presence of `id_token_encrypted_response_*` parameters in the SIOP request.
- [ ] TBD - Interactions between verifier and issuer. Do verifiers call an issuer's well-known did config?

jmandel:
Thanks for these! Can you review https://github.com/smart-on-fhir/health-cards/pull/38 to ensure I've covered (1-3) correctly? And re: (4) this is a broader topic that needs more explicit language in the spec; would you mind capturing this in a separate issue?

madaster97:
Hey Josh,

1 and 2 look good! But I'm realizing I had 3 wrong.

The `id_token_encrypted_response_*` value is presented by the RP (lab/issuer) and used by the SIOP (holder) to encrypt the **SIOP response**, and not used by the issuer to encrypt VCs.

The way that VCs get encrypted is not well-defined. During a SIOP flow, an issuer just sort of _notices_ that a holder has encryption key(s), and chooses to use one at it's own discretion. I think that sort of explanation should replace what you have in the VCs section.

I don't think we should say too much on the topic, other than mentioning that RPs can choose to not trust a given holder if it doesn't own/present an encryption key. In the `$HealthConnect` workflow, this is managed using the `encryptForKeyId` parameter, but could also optionally be enforced by an issuer by just choosing any key the wallet has.

[...]

PR #38, opened 2021-01-21
Clarifications for #29

jmandel:


jmandel:
OK @madaster97, this is ready for another look.

docs/index.md
```diff
@@ -400,13 +400,13 @@ See [Modeling Verifiable Credentials in FHIR](./credential-modeling/) for detail
...
+In this step, the user learns that new lab results are available (e.g., by receiving a text message or email notification). To facilitate this workflow, the lab can include a link to help the user download the credentials directly, e.g., from at a login-protected page in the Lab's patient portal. The file should be served with a `.fhir-backed-vc` file extension, so the Health Wallet app can be configured to recognize this extension. Contents should be a JSON object containing an array of Verifiable Credential JWTs, which MAY be either JWE or JWS, at the issuer's discretion. (In the case where the user has not connected a wallet to the issuer in advance, these will necessarily be JWS values, since no encryption key is known; in the case where the user has connected a health wallet to the issuer, the issuer MAY choose to encrypt the `.fhir-backed-vc` file using a key from the health wallet's registered DID):
```
p2-apple, line 403:
```suggestion
In this step, the user learns that new lab results are available (e.g., by receiving a text message or email notification). To facilitate this workflow, the lab can include a link to help the user download the credentials directly, e.g., from at a login-protected page in the Lab's patient portal. The file should be served with a `.fhir-backed-vc` file extension, so the Health Wallet app can be configured to recognize this extension. Contents should be a JSON object containing an array of Verifiable Credential JWTs, which MAY be either JWE or JWS, at the issuer's discretion:

- in the case where the user has NOT connected a wallet to the issuer in advance, these will necessarily be JWS values, since no encryption key is known
- in the case where the user has connected a health wallet to the issuer, the issuer MAY choose to encrypt the `.fhir-backed-vc` file using a key from the Health Wallet's registered DID
```

p2-apple: One suggestion since I believe these call-outs are important enough to stand on their own.

madaster97:
@jmandel changes look good to me. Including @p2-apple 's additions.

PR #50, opened 2021-01-28
Consider QR Embedding for paper-based  presentation

jmandel:
This is a different point in the trade-off space from #49. 

Advantages: 

* No need to host (even encrypted) data on the web
* No need for verifier to be online

Disadvantages:
* QR codes are large, making scanning more difficult
* Different processing rules apply to "QR-Ready Health Cards" 
  * Issuers need to understand and support in a specific way
  * Verifiers need to understand and support in a specific way 
* Some content must be stripped from the VC to shrink payload sizes
* Some payloads just might not fit (but for COVID-19 Vaccination Credentials, this should not be a blocking issue)

Discussion questions: If we adopt this model, would we take some of the processing rules here and apply them to normal Health Cards? For example, including only short-form DIDs plus optionally issuer domains, and expecting verifiers to resolve long-form DIDs either from a published trust bundle or from the verifier's `.well-known/did-configuration` endpoint?

 Based on discussion below, next steps are:
- [ ] Draft updates to this PR incorporating the suggestions [here ](https://github.com/smart-on-fhir/health-cards/pull/50#issuecomment-772072678)(@madaster97 would love your help)
- [ ] Provide guidance for keeping `Bundle.entry.fullUrl` values small in size and avoiding references to specific servers -- recommend `resource:1`, `resource:2`, etc as `fullUrl` values to support bundle-local resolution (they are short, internally unique, and avoid external correlation)
- [ ] Get feedback on usability of QR codes vs total size (when printed on paper, and when displayed on phone screens
- [ ] Ensure we're comfortable with the inclusion of a `zip` header in our JWS structures

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -664,15 +664,16 @@ In this flow, the user obtains a physical or virtual Health Card QR code. This c
 * `kid` in the JWS header is omitted
 * `sub` in the JWS payload is omitted
 * `iss` in the JWS payload is a short-form ion DID
+* `.vc.issuerDomain` (optional) is populated with the branded domain of the issuer
 * `.vc.credentialSubject.fhirBundle` in the JWS payload is:
   * created without `Resource.meta` elements
   * created without `Resource.text` elements
   * created without `Coding.display` elements
   * minified (i.e., all optional whitespace has been stripped)
  
-The *QR-Ready Health Card* is turned into a QR code by gzipping and then base64 encoding. If the resulting payload is larger than 4296 bytes, the issuer should regenerate the *QR-Ready Health Card* redacting FHIR resources until it is within limits.
+The *QR-Ready Health Card* is turned into a QR code by gzipping and then base64url encoding. If the resulting payload is larger than 4296 bytes, the issuer should refactor the resources into multiple *QR-Ready Health Cards*, populating `.vc.credentialSubject.fhirBundleSplit` with the number of cards in the set, and populating `.vc.credentialSubject.fhirBundleId` with the same UUID for each card in the set.
 
-In order to verify a Health Card QR Code, the verifier must be able to look up the long-form DID value (i.e., the initial state) associated with the issuer's short-form ion DID (e.g., by checking with a registry of known issuer DIDs).
+In order to verify a Health Card QR Code, the verifier must be able to look up the long-form DID value (i.e., the initial state) associated with the issuer's short-form ion DID (e.g., by checking with a registry of known issuer DIDs, or by checking the `.well-known/did-configuration` associated with the `.vc.issuerDomain`).
 
 
 ## Potential Extensions
--- a/docs/index.md
+++ b/docs/index.md
@@ -669,9 +669,10 @@ In this flow, the user obtains a physical or virtual Health Card QR code. This c
   * created without `Resource.meta` elements
   * created without `Resource.text` elements
   * created without `Coding.display` elements
-  * minified (i.e., all optional whitespace has been stripped)
+* entire payload is minified (i.e., all optional whitespace has been stripped)
+* entire payload is gzipped before being signed (i.e., the signed bytes are the minified gzipped JSON payload object)
  
-The *QR-Ready Health Card* is turned into a QR code by gzipping and then base64url encoding. If the resulting payload is larger than 4296 bytes, the issuer should refactor the resources into multiple *QR-Ready Health Cards*, populating `.vc.credentialSubject.fhirBundleSplit` with the number of cards in the set, and populating `.vc.credentialSubject.fhirBundleId` with the same UUID for each card in the set.
+The *QR-Ready Health Card* as a JWT is directly used as the text body of a QR code. If the resulting payload is larger than 4296 bytes, the issuer should refactor the resources into multiple *QR-Ready Health Cards*, populating `.vc.credentialSubject.fhirBundleSplit` with the number of cards in the set, and populating `.vc.credentialSubject.fhirBundleSet` with the same UUID for each card in the set.
 
 In order to verify a Health Card QR Code, the verifier must be able to look up the long-form DID value (i.e., the initial state) associated with the issuer's short-form ion DID (e.g., by checking with a registry of known issuer DIDs, or by checking the `.well-known/did-configuration` associated with the `.vc.issuerDomain`).
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -661,6 +661,9 @@ The process begins with a QR code or `openid://` link. The only differences are:
 
 In this flow, the user obtains a physical or virtual Health Card QR code. This code can be shared with verifying parties or loaded into a new wallet app by scanning the QR code. The Health Card QR code is created from a specially-issued *QR-Ready Health Card* with the following constraints applied:
 
+* payload is minified (i.e., all optional whitespace has been stripped)
+* payload is compressed with the DEFLATE (RFC1951) algorithm before being signed
+* `zip: "DEF"` is specified in the JWS header
 * `kid` in the JWS header is omitted
 * `sub` in the JWS payload is omitted
 * `iss` in the JWS payload is a short-form ion DID
@@ -669,8 +672,6 @@ In this flow, the user obtains a physical or virtual Health Card QR code. This c
   * created without `Resource.meta` elements
   * created without `Resource.text` elements
   * created without `Coding.display` elements
-* entire payload is minified (i.e., all optional whitespace has been stripped)
-* entire payload is gzipped before being signed (i.e., the signed bytes are the minified gzipped JSON payload object)
  
 The *QR-Ready Health Card* as a JWT is directly used as the text body of a QR code. If the resulting payload is larger than 4296 bytes, the issuer should refactor the resources into multiple *QR-Ready Health Cards*, populating `.vc.credentialSubject.fhirBundleSplit` with the number of cards in the set, and populating `.vc.credentialSubject.fhirBundleSet` with the same UUID for each card in the set.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -651,20 +651,35 @@ The process begins with a QR code or `openid://` link. The only differences are:
     ```
 ---
 
+### Paper-based fallback: Health Card QR Codes
+* Allows basic use of health cards for users without a smartphone
+* Allow smartphone-enabled users to print a usable backup
+* Allows near-complete health card contents to be shared with a verifier
+* Requires that the verifier has pre-configured a set of known issuers
+* Does not capture a digital record of a signed request (no DID-SIOP Request flow)
+* Does not capture a digital record of the presentation  (no Verifiable Presentation flow)
+
+In this flow, the user obtains a physical or virtual Health Card QR code. This code can be shared with verifying parties or loaded into a new wallet app by scanning the QR code. The Health Card QR code is created from a specially-issued *QR-Ready Health Card* with the following constraints applied:
+
+* `kid` in the JWS header is omitted
+* `sub` in the JWS payload is omitted
+* `iss` in the JWS payload is a short-form ion DID
+* `.vc.credentialSubject.fhirBundle` in the JWS payload is:
+  * created without `Resource.meta` elements
+  * created without `Resource.text` elements
+  * created without `Coding.display` elements
+  * minified (i.e., all optional whitespace has been stripped)
+ 
+The *QR-Ready Health Card* is turned into a QR code by gzipping and then base64 encoding. If the resulting payload is larger than 4296 bytes, the issuer should regenerate the *QR-Ready Health Card* redacting FHIR resources until it is within limits.
+
+In order to verify a Health Card QR Code, the verifier must be able to look up the long-form DID value (i.e., the initial state) associated with the issuer's short-form ion DID (e.g., by checking with a registry of known issuer DIDs).
+
+
 ## Potential Extensions
 
 ### Fallback for smartphone-based offline presentation
 
 We should be able to specify additional "return paths" in the DID SIOP workflow that don't depend on an HTTP upload but instead rely on local transfer (e.g., via NFC or bluetooth)
-
-### Fallback for users without a smartphone
-
-While it's hard to provide the same level of functionality and convenience without a mobile phone, there are still steps we can take to allow broader use of these verifiable credentials. Here's one possibleS approach to graceful degradation:
-
-* Lab generates VCs that aren't bound to any specific user DID
-* Lab makes VCs available for download
-* User prints a QR Code conveying the VC, or a link to a hosted copy of the VC (optionally protected by a password or PIN)
-* Verifier scans the barcode, retrieves the VC, and verifies signatures -- then relies on out-of band relationship with the user to match the VC to a real-world identity. For example, the user may be an employee or customer of the verifier, and thus the user's name and phone number may be known by the verifier in advance. The verifier must compare the identity attributes inside the VC with the attributes they have verified out of band.
 
 
 


jmandel:
Working from a life-size example `fhirBundle`, this gets us in range -- though the `zip: "DEF"` is really important to the story, and it's a non-standardized header for JWT (exists in JWS spec).

```
With zip:DEF:  JWS is 1505 bytes
eyJraWQiOiJkaWQ6aW9uOkVpQmQxVHg3OS1iMmtseG1ucHF3SUI5X3ZiUVFja0ltMjdMZWVJQ0hFTHBaZHcjc2lnbmluZy1rZXktMSIsInppcCI6IkRFRiIsImFsZyI6IkVTMjU2In0.eJzVVG1v2zYQ_isB-3GmRFJSLOtTk7RAM6RDOqf70CEYKJKy2EiiJlJ-WeD_viNlJ-7WBigwoJgBASZ5d8_dc8_dI3otTOfU1qHid1Q719sijjebTbRJIjOsYkZoHotBSdU5zRsbrym6nyG369Wph2354GrFG1dHgg_SvhJmrSVdoNkLNtMB-8OLdrptx07_xZ02nUfX1o5qQAWSWhZwV7zVl5LebecLXLKHZtt2_Z-b68vFH-vywwfxcN2y-Y1S11fv3t70n-QGTRF4J9Qb7qAOxAgjmJxjQu9oXmSLgpCIEPIJLJ9rX47lZyWAqEekJTjBY1Xr4Tc1WJ9XgdKIRPRwezl2slHedlDWjINQd4EydHg4UoiEaRqI6iPMEAANO-D1EVVj03wcGjAYh64YRyi0ZCqVhFWYzEWC00xRXNIqwTRjpOQ5LTMqIMYR7yvYt8AgQHigrVPdlDaAjQHoyP9Ku3osI2Ha-G7gne35AE7vQkPinBB8nuAEu2G0Djr18OoW4q15M6orIwMKQ3vfpEBbpX2jPIb1byPQ3jwV_wjVS92tgoHdWafaQx6QhlNDqzvTmNUuqpt5UKNHWAa7eM0wNC3xHZpg3_8KsPvZv-LUxvba8SZSW972jfKBjhmDVUtZEvLteKsm6nmrG-gCuuggy8GGzqz0Wnmy0M-m9ufLCN17r5XqZJBiy0NXSw3aPaiKLjIKksKMwAOXErphA0KjuzA9AH32nuvubOm8rIV2B1hnNh6kNxYSP9AKxmkWyh0nlaCPy5C4H2DudQmRnQIxmfZLQvvadOqk5CzL8OHzAafO1AbK389O3FTLdXPi9hkqj_iRk9dHOj0cEH-_985fkW3ClExVtsB0njKc5mWO-ZylmM35PFVcqZSVL8v2-nT8ocGOu9GG0fEJOCVDkkIAqxNXL-rqqCU_pbHVMhbr7bOKGJlPMuoPsxLSqRSMgDgZoXgpVMcHbehxpsDFCDEOwdAL4E63h9USROBXCy3SrEiSn-AP8ZroBw0TtFseCoeRUjPUGDFV-g_km8P9M_TxxmM3xv0ytmWQ4o1xZ69I-NGAYpyBNXPR940GsgIpatDKvjFW2VtjtQN1X_timReUfV50pwl83yLaf1MPXC0qCmskZVWKU0JTzM8Zw5nMZMoXKs-r5Lv0EJbxEymnj-x_qxa2-BFqmX9DLRJ0Mtl_KZbZj9SRWDAqy0ximUsQk0goznnCMSHsvBIkq9JcvqyjJ0aOyx9dXF6d3dZ8aLnYof-uANiN-78BdjIWpA.z9rD7TT6JJmtIP29J37lMfXfjmJ-JG24Q59DKy5NQxa-oT-YaS10SuwpaPGXPB72qt-eDCyfCsCAWQX60T3LYg
```

```
Without zip:DEF, JWS is 3425 bytes
eyJraWQiOiJkaWQ6aW9uOkVpQmQxVHg3OS1iMmtseG1ucHF3SUI5X3ZiUVFja0ltMjdMZWVJQ0hFTHBaZHcjc2lnbmluZy1rZXktMSIsImFsZyI6IkVTMjU2In0.eyJAY29udGV4dCI6WyJodHRwczovL3d3dy53My5vcmcvMjAxOC9jcmVkZW50aWFscy92MSJdLCJ0eXBlIjpbImh0dHBzOi8vc21hcnRoZWFsdGguY2FyZHMjY292aWQxOSIsImh0dHBzOi8vc21hcnRoZWFsdGguY2FyZHMjaGVhbHRoLWNhcmQiLCJodHRwczovL3NtYXJ0aGVhbHRoLmNhcmRzI2ltbXVuaXphdGlvbiJdLCJpc3N1ZXIiOiJkaWQ6aW9uOkVpQmQxVHg3OS1iMmtseG1ucHF3SUI5X3ZiUVFja0ltMjdMZWVJQ0hFTHBaZHciLCJpc3N1YW5jZURhdGUiOiIyMDIwLTA2LTAxVDE4OjU5OjAwLjAwMFoiLCJjcmVkZW50aWFsU3ViamVjdCI6eyJpZCI6IiIsImZoaXJWZXJzaW9uIjoiNC4wLjEiLCJmaGlyQnVuZGxlIjp7InJlc291cmNlVHlwZSI6IkJ1bmRsZSIsInR5cGUiOiJjb2xsZWN0aW9uIiwiZW50cnkiOlt7ImZ1bGxVcmwiOiJ1cm46dXVpZDpiMmU0ZDAyZi0wN2MzLTQ1ZTEtYjFmMy0xNTIwYmE4MWI1MWMiLCJyZXNvdXJjZSI6eyJyZXNvdXJjZVR5cGUiOiJQYXRpZW50IiwiZXh0ZW5zaW9uIjpbeyJ1cmwiOiJodHRwczovL2dpdGh1Yi5jb20vVHJhbnNwYXJlbnRIZWFsdGgvODAwLTYzLTMtdHJ1c3RtYXJrI1AiLCJ2YWx1ZUNvZGUiOiJQMiJ9XSwiaWRlbnRpZmllciI6W3sidXNlIjoidXN1YWwiLCJ0eXBlIjp7ImNvZGluZyI6W3sic3lzdGVtIjoiaHR0cDovL3Rlcm1pbm9sb2d5LmhsNy5vcmcvQ29kZVN5c3RlbS92Mi0wMjAzIiwiY29kZSI6Ik1SIn1dfSwic3lzdGVtIjoiaHR0cDovL2hvc3BpdGFsLmV4YW1wbGUub3JnIiwidmFsdWUiOiJtMTIzIn1dLCJuYW1lIjpbeyJmYW1pbHkiOiJBbnlwZXJzb24iLCJnaXZlbiI6WyJKb2huIiwiQi4iXX1dLCJnZW5kZXIiOiJtYWxlIiwiYmlydGhEYXRlIjoiMTk1MS0wMS0yMCIsImFkZHJlc3MiOlt7ImxpbmUiOlsiMTIzIE1haW4gU3QiXSwiY2l0eSI6IkFueXRvd24iLCJwb3N0YWxDb2RlIjoiMTIzNDUiLCJjb3VudHJ5IjoiVVMifV0sImNvbnRhY3QiOlt7InRlbGVjb20iOlt7InN5c3RlbSI6InBob25lIiwidmFsdWUiOiI1NTUtNTU1LTU1NTUiLCJ1c2UiOiJob21lIn0seyJzeXN0ZW0iOiJlbWFpbCIsInZhbHVlIjoiam9obi5hbnlwZXJzb25AZXhhbXBsZS5jb20ifV19XX19LHsiZnVsbFVybCI6InVybjp1dWlkOjMyZWQ0ZTU5LTE3NDItNDhiOC1hNzI0LTI3YTc0ZWFlZTQyYiIsInJlc291cmNlIjp7InJlc291cmNlVHlwZSI6IkltbXVuaXphdGlvbiIsInN0YXR1cyI6ImNvbXBsZXRlZCIsInZhY2NpbmVDb2RlIjp7ImNvZGluZyI6W3sic3lzdGVtIjoiaHR0cDovL2hsNy5vcmcvZmhpci9zaWQvY3Z4IiwiY29kZSI6IjIwNyJ9XX0sInBhdGllbnQiOnsicmVmZXJlbmNlIjoiUGF0aWVudC9TY2VuYXJpbzFQYXRpZW50In0sIm9jY3VycmVuY2VEYXRlVGltZSI6IjIwMjEtMDEtMDFUMTE6NDU6MzMrMTE6MDAiLCJwcmltYXJ5U291cmNlIjp0cnVlLCJsb2NhdGlvbiI6eyJyZWZlcmVuY2UiOiJMb2NhdGlvbi9TY2VuYXJpbzFMb2NhdGlvbiJ9LCJsb3ROdW1iZXIiOiJMb3QgIzAwMDAwMDEiLCJwcm90b2NvbEFwcGxpZWQiOlt7InNlcmllc0Rvc2VzUG9zaXRpdmVJbnQiOjJ9XSwic3ViamVjdCI6eyJyZWZlcmVuY2UiOiJ1cm46dXVpZDpiMmU0ZDAyZi0wN2MzLTQ1ZTEtYjFmMy0xNTIwYmE4MWI1MWMifX19LHsiZnVsbFVybCI6InVybjp1dWlkOjNhZTlmMTAwLTQyZjQtNDAxNC1hNjIyLTVkNWQ0YTllODhmMyIsInJlc291cmNlIjp7InJlc291cmNlVHlwZSI6IkltbXVuaXphdGlvbiIsImlkIjoiU2NlbmFyaW8xSW1tdW5pemF0aW9uMiIsInN0YXR1cyI6ImNvbXBsZXRlZCIsInZhY2NpbmVDb2RlIjp7ImNvZGluZyI6W3sic3lzdGVtIjoiaHR0cDovL2hsNy5vcmcvZmhpci9zaWQvY3Z4IiwiY29kZSI6IjIwNyJ9XX0sInBhdGllbnQiOnsicmVmZXJlbmNlIjoiUGF0aWVudC9TY2VuYXJpbzFQYXRpZW50In0sIm9jY3VycmVuY2VEYXRlVGltZSI6IjIwMjEtMDEtMjlUMTE6NDU6MzMrMTE6MDAiLCJwcmltYXJ5U291cmNlIjp0cnVlLCJsb2NhdGlvbiI6eyJyZWZlcmVuY2UiOiJMb2NhdGlvbi9TY2VuYXJpbzFMb2NhdGlvbiJ9LCJsb3ROdW1iZXIiOiJMb3QgIzAwMDAwMDciLCJwcm90b2NvbEFwcGxpZWQiOlt7ImRvc2VOdW1iZXJQb3NpdGl2ZUludCI6Miwic2VyaWVzRG9zZXNQb3NpdGl2ZUludCI6Mn1dLCJzdWJqZWN0Ijp7InJlZmVyZW5jZSI6InVybjp1dWlkOmIyZTRkMDJmLTA3YzMtNDVlMS1iMWYzLTE1MjBiYTgxYjUxYyJ9fX0seyJmdWxsVXJsIjoidXJuOnV1aWQ6YzkyMWRiNWQtZDhkMC00YzMxLThhM2EtMDAyNmZjMDVmNDhkIiwicmVzb3VyY2UiOnsicmVzb3VyY2VUeXBlIjoiTG9jYXRpb24iLCJuYW1lIjoiQUJDIFBoYXJtYWN5Iiwic3ViamVjdCI6eyJyZWZlcmVuY2UiOiJ1cm46dXVpZDpiMmU0ZDAyZi0wN2MzLTQ1ZTEtYjFmMy0xNTIwYmE4MWI1MWMifX19XX19fQ.zsHx-gDqo7l5n9RJqJc4ejrGRUWSzw-8Ltgq0kwb2XLX-FqoiJjzVomSjz6eW1LO362SMXXNdYja4FpXvTQHfw

gillesroux:
Josh, thanks for the proposal! Here are a few initial comments/questions:

**Digital signature**: With the recommendation to omit ‘kid’, how will the reader know what key/certificate to verify the signature with? Could we use a ‘x5u’ parameter to point to a URL hosting the full signing certificate chain. This would allow supporting several signing issuers and giving them the ability to issue sub-certificates (eg: pharmacy branches). One downside is that the reader would need to be online, but this could be mitigated by caching or even-precaching common certificate chains.

**Payload content**: If I understand well, for the case of an immunization proof, the proposal is to use a VC with a FHIR bundle containing a Patient and a set of Immunization resources, is that correct? Do you know if verifiers are well equipped to evaluate the immunization status based on the established immunization protocols (eg: the min/max interval allowable between two doses for each vaccine type)? Another approach would be to use an ImmunizationRecommendation to communicate whether the vaccination is complete or not. This means the immunization evaluation would fall on the issuer rather than the verifier. Another thing to consider is whether verifiers have specific conditions requiring access to information about individual vaccination events.

**Payload size**: The process of trimming the FHIR bundle until the JWS fits in a QR code seems tricky and makes it hard to give predictability on the information being conveyed. Do you think it would be workable to mandate the fields to be included (eg: required FHIR fields) and maybe even flatten the resulting structure. This approach is of course less flexible but more predictable and yields much smaller QR codes.

**Fallback for smartphone-based offline presentation**: Using NFC/BLE for data transfer seems workable but introduces points of failures, for example devices with BT off, lack of support from readers… A simple alternative would be to rely on the paper-based fallback QR code.

jmandel:
Thanks for the close read, and the questions and comments.

>Digital signature: With the recommendation to omit ‘kid’, how will the reader know what key/certificate to verify the signature with? Could we use a ‘x5u’ parameter to point to a URL hosting the full signing certificate chain. This would allow supporting several signing issuers and giving them the ability to issue sub-certificates (eg: pharmacy branches). One downside is that the reader would need to be online, but this could be mitigated by caching or even-precaching common certificate chains.

​The rationale for omitting `kid`​ was just that the DID is already present inside the payload, and I wanted to avoid duplication in the header. Maybe I should just say: `kid`​ is present and valued as a short-form DID + `#fragment`, to avoid this confusion, at the cost of a small amount of repetition. (Also, without knowing the `#fragment`, technically a verifier would need to iterate through the short list -- usually just a single one-- of signing keys in the DID doc to validate.) Note that re: `x5u`, we don't actually use certificate chains in the Health Cards spec (just signing keys in the form of "bare" JWKs discoverable in DID).


>Payload content: If I understand well, for the case of an immunization proof, the proposal is to use a VC with a FHIR bundle containing a Patient and a set of Immunization resources, is that correct? Do you know if verifiers are well equipped to evaluate the immunization status based on the established immunization protocols (eg: the min/max interval allowable between two doses for each vaccine type)?

​The messy clinical/scientific reality is that this will be a moving target. (That's the kind of thing my [`summary.ts` example](https://github.com/microsoft-healthcare-madison/health-wallet-demo/commit/ef3dcc1db5d6d7639f5b555a63b12fba65ec1429) was looking at.) As a community, I think we need to develop good data-driven open source tools for capturing these summarization algorithms, but they can't be "pre-calculated" because the rules will keep changing.

>Another approach would be to use an ImmunizationRecommendation to communicate whether the vaccination is complete or not. This means the immunization evaluation would fall on the issuer rather than the verifier. Another thing to consider is whether verifiers have specific conditions requiring access to information about individual vaccination events.

​I understand where you're coming from. I don't think we want to rely on the issuer to get this right, especially since:
* our understanding can change in both directions (learning that a vaccine works better or worse than expected); there can be many conclusions beyond "fully vaccinated" and "not vaccinated"; and 
* as you suggest, different verifiers will need to apply different rules, especially for different jurisdictions (e.g.,  we won't have specific inter-dose time intervals that the whole world agrees on; jurisdictions will set their own rules, and some will lead or lag.)

>Payload size: The process of trimming the FHIR bundle until the JWS fits in a QR code seems tricky and makes it hard to give predictability on the information being conveyed.

​I agree it's tricky. I think there's a way to thread the needle while remaining general-purpose (i.e., don't bake in assumptions based on the particular kind of Health Card); expressive (allow for flexibility of the FHIR data model); and small enough (keep the result within a single QR code for all "normal" cases, and define an optional fallback if multiple QRs are needed).

> Do you think it would be workable to mandate the fields to be included (eg: required FHIR fields) and maybe even flatten the resulting structure. This approach is of course less flexible but more predictable and yields much smaller QR codes.

We definitely will mandate fields at the level of the FHIR profiles, like the [Vaccination Credential FHIR Profiles](https://github.com/dvci/vaccine-credential-ig) do now, between `mustSupport` and minimum cardinalities. We may even prohibit fields, but right now the paradigm is "discourage anything that isn't `mustSupport`", and that could be strong enough language.  ​I'm open to further tricks like flattening if we get stuck; but it'd be ideal to avoid significant changes the task that versifiers have to perform.

>Fallback for smartphone-based offline presentation: Using NFC/BLE for data transfer seems workable but introduces points of failures, for example devices with BT off, lack of support from readers… A simple alternative would be to rely on the paper-based fallback QR code.

​Indeed -- this language was a nod toward some future exploration, but it's definitely out of scope for the time being (unlike paper-based QR codes of some kind, which are a must-do). Might be cleaner to cut this altogether.

agropper:
Are all of these compromises are due to the assumption that a credential has to be on a mobile device in order to be patient-controlled? 

If each patient had a custodial wallet (similar to [IndiaStack](https://www.indiastack.org) and the [NHDM](https://government.economictimes.indiatimes.com/news/digital-india/delhi-plans-cloud-based-health-management-system-to-rollout-qr-code-based-health-cards-soon/80386505)) then the mobile device would be just an optional authenticator and policy editor having nothing specific to healthcare or pandemics. Paper credentials, including QR codes, would simply point issuers and verifiers to the custodial wallet in the cloud.

jmandel:
Interesting @agropper -- I think you may be describing a new point on the trade-off curve. So far we have identified options #49 and #50, and would like to identify any relevant alternatives (as well as improvements to these two approaches).

Would you be willing to review #49 and explain the key differences to the "custodial wallet" concept? In comments on that issue, it would be good to describe how a custodial wallet is different from a hosted credential that you can redact (take offline) when you choose. Best if you can include a workflow description; any assumptions in terms of available infrastructure; and trade-offs with respect to user experience, privacy, or cost.

(Reading through the links you shared, it seems like the card systems basically encode a patient identifier within a web accessible health record, or geocoordinates for a patient's home, so that (presumably?) an already authorized healthcare worker would then know which account to open up during a visit. Apologies if I'm misunderstanding.)

arztnh:
On 1/29/2021 5:46 PM, Josh Mandel
      wrote:
    
    
      
        Payload content: If I understand well, for the case of an
          immunization proof, the proposal is to use a VC with a FHIR
          bundle containing a Patient and a set of Immunization
          resources, is that correct? Do you know if verifiers are well
          equipped to evaluate the immunization status based on the
          established immunization protocols (eg: the min/max interval
          allowable between two doses for each vaccine type)?
      
      ​The messy clinical/scientific reality is that this will be a
        moving target. (That's the kind of thing my summary.ts example was
        looking at.) As a community, I think we need to develop good
        data-driven open source tools for capturing these summarization
        algorithms, but they can't be "pre-calculated" because the rules
        will keep changing.
      
        Another approach would be to use an
          ImmunizationRecommendation to communicate whether the
          vaccination is complete or not. This means the immunization
          evaluation would fall on the issuer rather than the verifier.
          Another thing to consider is whether verifiers have specific
          conditions requiring access to information about individual
          vaccination events.
      
      ​I understand where you're coming from. I don't think we want
        to rely on the issuer to get this right, especially since:
      
        our understanding can change in both directions (learning
          that a vaccine works better or worse than expected); there can
          be many conclusions beyond "fully vaccinated" and "not
          vaccinated"; and
        as you suggest, different verifiers will need to apply
          different rules, especially for different jurisdictions (e.g.,
          we won't have specific inter-dose time intervals that the
          whole world agrees on; jurisdictions will set their own rules,
          and some will lead or lag.)
      
    
    
    This issue is tricky. It probably makes sense for the issuer to send
    ImmunizationRecommendation if they know it, and to specify the
    jurisdiction to which it applies. A verifier would then be free to
    use it or ignore it. The reality in the field is that many (dare I
    say most) verifiers will not have local clinical decision support to
    generate local recommendations. With COVID vaccines especially I
    think there will be few to no jurisdictional differences as everyone
    will likely follow the clinical guidelines passed on from the
    manufacturers and so far validated by the respective CDCs in each
    jurisdiction. But even if the recommendations differ better to have
    the OPTION of getting them along with the immunization history.
    
    Noam

jmandel:
@arztnh thank you for the perspective on this. I'd maybe raise this as a [question on the Immunizations IG](https://github.com/dvci/vaccine-credential-ig/issues) -- specifically, whether we should model an `ImmunizationRecommendation` as an optional bundle element. I lean toward "no" for a few reasons (I should emphasize that none of these reasons strictly rule out this approach; it's just that they add up, for me, to a compelling conclusion):

* `ImmunizationRecommendation` is a FHIR Maturity Model level 1 and we don't have much experience with it
* `ImmunizationRecommendation` is not part of FHIR US Core, and we're trying to build on US core where possible 
* While immunization registries may be in a position to generate these recommendations, other issuers likely won't
* The recommendations are going to be *very* hard to interpret in light of changing CDS algorithms, against the backdrop of new variants, new vaccine products, and new studies
* In many cases, packing a recommendation at a single point in time could be misleading (e.g., a Health Card showing you have received one dose of the Moderna vaccine, and also including an `ImmunizationRecommendation` indicating you're not yet fully vaccinated -- if you present this alongside a second health card from your local clinic showing that you've received the final dose, but the clinic rather than the IIS created this second health card and it lacks any `ImmunizationRecommendation`, then the job of a verifier gets even harder, having to determine which Recommendations are stale in the context of new data)
* Verifiers are going to have to implement logic to assess health cards and draw conclusions, no matter whether Health Cards include `Immunization`s`, `ImmunizationRecommendation`s, or some combination; we'll need to support this with open source libraries and tools to make it easier
* Issuers will sometimes have only "one chance" to capture relevant details; so we should prefer to focus on immutable facts rather than on the outputs of algorithms that could be evaluated later

arztnh:
On 1/31/2021 9:31 AM, Josh Mandel
      wrote:
    
    
      
        ImmunizationRecommendation is a FHIR Maturity
          Model level 1 and we don't have much experience with it
      
    
    
    Active work is being done now.
    
    
      
        ImmunizationRecommendation is not part of FHIR
          US Core, and we're trying to build on US core where possible
      
    
    
    Fair enough. Not something I track closely.
    
    
      
        While immunization registries may be in a position to
          generate these recommendations, other issuers likely won't
      
    
    
    Which is an argument for it being optional which is all I think I
    was asking.
    
    
      
        The recommendations are going to be very hard to
          interpret in light of changing CDS algorithms, against the
          backdrop of new variants, new vaccine products, and new
          studies
      
    
    
    It's not that they are hard to interpret (if done  correctly)...
    
    
      
        In many cases, packing a recommendation at a single point in
          time could be misleading (e.g., a Health Card showing you have
          received one dose of the Moderna vaccine, and also including
          an ImmunizationRecommendation indicating you're
          not yet fully vaccinated -- if you present this alongside a
          second health card from your local clinic showing that you've
          received the final dose, but the clinic rather than the IIS
          created this second health card and it lacks any ImmunizationRecommendation,
          then the job of a verifier gets even harder, having to
          determine which Recommendations are stale in the context of
          new data)
      
    
    
    ...it's this point about them only being correct at a moment in
    time. But that is the inherent problem with immunizations (and
    future immunization needs) in general. You run a greater risk that
    the verifier won't have the insight or ability to assess the
    immunization history at all.
    
    
      
        Verifiers are going to have to implement logic to assess
          health cards and draw conclusions, no matter whether Health
          Cards include Immunizations, ImmunizationRecommendation`s,
          or some combination; we'll need to support this with open
          source libraries and tools to make it easier
      
    
    
    Ah, we have one of those: Our open source Immunization Calculation
    Engine (ICE) currently has rules for the US and we'd love to find
    some sponsorship to expand those rules to other parts of the world
    including the WHO
      schedule. See
    https://cdsframework.atlassian.net/wiki/spaces/ICE/overview
    
    
      
        Issuers will sometimes have only "one chance" to capture
          relevant details; so we should prefer to focus on immutable
          facts rather than on the outputs of algorithms that could be
          evaluated later
      
    
    
    As long as those "immutable facts" (see an earlier comment on
    whether an immunization record is really immutable) can be
    put to proper use.
    
    Thanks,
    
    Noam

agropper:
The issue of clinical / public health verification context is not my area of expertise but it's a cousin to revocation, in general. 

VCs create a revocation problem because they're designed to avoid verifiers "calling home" to the issuer for privacy reasons. This is critical for drivers licenses and lifelong educational credentials but it may be overkill for vaccination in a pandemic.

So, I agree with @jmandel that issuers should stick to the facts they know and avoid interpretation. The verifiers, however, should have a means to incorporate context into their process. I'm not sure there's anything much the issuers can do to help the verifiers also check a context. This issuers are not in a position to "revoke" their credential in the classical sense. 

I would be more interested to build in some audit capability to avoid "verifier shopping" the way we do prescription drug monitoring programs (PDMP). These context providers are a giant privacy problem when used for behavioral health but they might be well tolerated in a pandemic where public health context is economically valuable and constantly changing.

gillesroux:
> ​The rationale for omitting `kid`​ was just that the DID is already present inside the payload, and I wanted to avoid duplication in the header. Maybe I should just say: `kid`​ is present and valued as a short-form DID + `#fragment`, to avoid this confusion, at the cost of a small amount of repetition. (Also, without knowing the `#fragment`, technically a verifier would need to iterate through the short list -- usually just a single one-- of signing keys in the DID doc to validate.) Note that re: `x5u`, we don't actually use certificate chains in the Health Cards spec (just signing keys in the form of "bare" JWKs discoverable in DID).

Thanks for pointing this out, I had missed that the public key was obtainable from the issuer did. Can you confirm that we're talking about the 'iss' field of the verifiable credential? Is there a mechanism for signing the key by another issuer all the way to a trusted root? I'm thinking of the case where the issuer might be a small lab that is not known to the verifier, or where a large lab/pharmacy decides to use "sub-issuers" to identify individual chains or practitioners.

Regarding the 'kid' field, I'm not concerned about omitting the did if we clearly specify where it can be found, however I would prefer not having to iterate over the different keys. An alternative could be to set 'kid' to just the fragment of the credential issuer from the payload.

jmandel:
>Can you confirm that we're talking about the 'iss' field of the claim? 

Yeah, in an issued VC, the JWS header's `kid` identifies a specific key, and the payloads `iss` is a DID resolvable to the full DID doc in which that key is defined.

> Is there a mechanism for signing the key by another issuer all the way to a trusted root? I'm thinking of the case where the issuer might be a small lab that is not known to the verifier, or where a large lab/pharmacy decides to use "sub-issuers" to identify individual chains or practitioners.

Agreed that mapping these kinds of org structures can be useful; it's not something we've attempted to specify in the Health Cards framework. I'll note that it is something that could be layered on in the trust framework, but we've been focused on "list of issuers and a flat list of DIDs for each issuer" to start. 


> An alternative could be to set 'kid' to just the fragment of the credential issuer from the payload.
👍

gillesroux:
As a followup, how do you envision the signature validation process to take place?

I could imagine the following:
 1. Verifier checks that iss is in a pre-configure allow-list of trusted issuers
 2. Verifier resolves the short-form did in iss
 3. Verifier checks the signature using the public key at iss#kid in the did document

Is the did resolution using ion fast enough to be done in real-time during verification? Would it be viable to pre-cache the resolved documents for all dids in the allow-list?

jmandel:
That all correct. Re: signature validation, (2) is the only step that might be slow, and I'm basically working from the assumption that verifiers need a cached list of full issuer DIDs for this scheme to work. Up to this point, we've said "only ever put long-form DIDs in a JWT", and we're only relaxing this for considerations of size. Using cached or simply decoded DID values right now is mainly a point about which runtime dependencies we want to take, rather than about latency: we say that in Health Cards v1 we won't depend on any DIDs actually being resolved against an ion node.)

madaster97:
It sounds like QR embedding is the way to go. I have a recommendation for what we should profile in the spec based on conversations so far. 

## Assumptions
- Verifiers will publish DID documents on their `.well-known/did-configuration` endpoint
- zip deflation is a reliable tool
- card splitting is a reliable tool
- QR version 40 codes are acceptable, with length limits of 4296 characters

## Proposed Spec Changes
### Addition of QR Presentation Workflow
We would be introducing a new workflow for verifiers, where the verifier would scan the QR code containing the embedded VCs. This would likely become the standard workflow, and the SIOP based workflows would be another optional workflow available.

Note the QR Presentation doesn't actually involve any verifiable presentations (VPs). While QR-ready VCs could be presented in a SIOP workflow as a VP, the workflow we're trying to enable is just for verifiers to scan a QR-ready VC.

### Define a QR-Ready VC
In general, we should aim for **all issued VCs** to also be compatible with QR-scanning. This way, a wallet app could choose to distribute these VCs via a SIOP workflow OR just presenting the VCs as QR codes.

In order to standardize QR-Ready VCs, issuers would need to follow these rules when generating any VC:
1. VCs will only reference the issuer with short-form DIDs
   - This assumes that we have a required mechanism for distributing DID documents for issuers.
   - If we are going to require issuers provide `did-configuration` endpoints, it makes sense that we also require that endpoint be present in the VCs as well.
   - **Note**: Short-form DIDs would only be used for referencing issuers. For wallets and verifiers, we would still assume long-form DIDs would be used.
2. If an issued VC is going to be greater than **4296 characters** (or a lower threshold tbd), an issuer MUST follow one or more of these schemes to make the VC compatible with QR presentation:
   - Issuer zip deflates the contents of the JWT, using the the zip: "deflate" JOSE header
   - Issuer separates the VC into multiple VCs, using the notation/markers defined above for linking cards ONLY if VCs have been split. This way, wallets and verifiers could function current-state, and enable this new split-card workflow when presented with split cards.

It really doesn't seem like we need much else to make this workflow viable, and that these changes wouldn't introduce many breaking changes to current issuers/verifiers.

### Other Points
On the JWE/JWS zip question. JWEs are only ever used to package up VCs, and would likely never be presented embedded in a QR code. So, we'd only ever ask that JWS payloads be zip deflated, since those are the only entities that may need to be QR distributed. An encrypted SIOP response would not itself be zipped, only the underlying VC claims within.

agropper:
Shouldn't we also consider a role for did:web?

jmandel:
Thanks @madaster97! I think your analysis is right on. From my perspective, next steps would be:

- [ ] Draft updates to this PR incorporating the suggestions here (@madaster97 would love your help)
- [ ] Get feedback on usability of QR codes vs total size (when printed on paper, and when displayed on phone screens
- [ ] Ensure we're comfortable with the inclusion of a non-standard `zip` header in our JWS structures

## Assumptions

> zip deflation is a reliable tool

Reliable yes but I'm not thrilled because `zip` isn't  a standardized JWS header feature. (I was thinking one way around this might be to do some QR-specific "repacking" of the JWS with a structure like `["<deatched JWS>", "<base64(deflate(payload))>"]`. This way a client could simply reconstruct a "normal" JWS from these two parts, and the resulting JWS wouldn't have any oddball compressed components. But it makes the issuer's job -- i.d., deciding when to split cards into sets -- harder because the issuer would need to actually do this compression speculatively to know if a card was too big to fit. I sure don't love that.)

> QR version 40 codes are acceptable, with length limits of 4296 characters

I'd really love to get some validation of this. Anybody aware of someone with domain experience (e.g., having conducting user testing) who could be invited to weigh in here? How easy are these to scan, and can they also be displayed on mobile phone screens effectively (in addition to paper)?


> Verifiers will publish DID documents on their .well-known/did-configuration endpoint

I'm missing where this assumption is invoked in your proposal. I suspect you meant "Issuers".


## Addition of QR Presentation Workflow

> This would likely become the standard workflow,

I'd wouldn't want to enshrine that "standard workflow" expectation in our spec (or even in our thinking); there are some key benefits to in-band, challenge/response based verifiable presentations. With the QR Presentation paradigm, we'd want to document the limitations clearly (e.g., no in-band way for the user to confirm the verifier's identity; no guided option for the user to select an appropriate, minimal set of credentials; no way to attach conditions or caveats to a presentation; no upgrade path to selective disclosure techniques).

To be clear, I _do_ totally agree that we'll want to explicitly add QR Presentation to the spec :-)

## Define a QR-Ready VC

> VCs will only reference the issuer with short-form DIDs

If we take this decision, Just to emphasize, populating `issuerDomain` (`issuerOrigin?`) in the VC payload will be important if we want to allow this chain of discovery from VC -> domain -> did-configuration doc.


## Other Points

> On the JWE/JWS zip question. JWEs are only ever used to package up VCs, and would likely never be presented embedded in a QR code. So, we'd only ever ask that JWS payloads be zip deflated, since those are the only entities that may need to be QR distributed. An encrypted SIOP response would not itself be zipped, only the underlying VC claims within.

This analysis looks right to me. To be clear though, the original point I raised was that while `zip` is defined in the JWE standard, it's *not* defined in the JWS standard -- meaning that it's not something that will be well supported out of the box by JWS libraries. I'd like to pay close attention to this in testing, if we go this route, to make sure this isn't too painful. (For the libraries I've looked at so far, hooking this into the plumbing has been pretty straightforward.)

jmandel:
I should add: even if we adopt these changes, the representation of a long-form Holder DID in QR codes is unsuitable (can't be used in a verifiable presentation anyway, and will increase the byte count beyond usable limits). So we'd still need an issuance time flag to control this aspect of generation for VCs intended for QR-embedding.

jmandel:
Really nice examples at https://github.com/minvws/nl-eHealth-experimental/tree/master/examples/syntetic-2 using protocol buffers to gain ~2x compression on these kinds of bundles. There's a challenge of agreeing on a "maximal"  set of protobuf definitions, because you can't just add in properties later the way you can with JSON. But compelling results re: size!

indigo99:
I did some research regarding the maximum scannable QR code size and found the following:

1. Minimum scannable size of a single QR code module is about 0.25mm
This figure depends on the scanner performance one one hand and the printer technology/dpi or display dpi on the other hand. 0.25mm is about the absolute minimum that we can expect to work across common devices. 
https://www.qrcode.com/en/howto/cell.html

2. A recommended guideline is module_count = 250 * scan_distance / qrcode_dimension
http://www.growthvine.com/blogs/kelly-flowers/2011/11/28/optimal-size-for-qr-codes-depend-on-a-few-things

If we assume the following:
 * QR code size: 35mm. This is the maximum that can be accommodated on a typical mobile device if we consider the required quiet zone. It's also reasonable for a printed QR code.
 * Error correction: L.

Then the maximum capacity of the QR code is:
 * Method 1: version 30 *137*137) -> 1700 bytes
 * Method 2: version 17 (85*85) -> 504 bytes

Since method 1 is really an upper bound, I would suggest to aim for a payload of ~500 bytes for the happy path scenario where the FHIR bundle follows the recommendations, and allow up to ~1500 bytes for degrade cases, with the understanding that the QR code may be harder to scan.

gillesroux:
I did some research regarding the maximum scannable QR code size and found the following:

1. Minimum scannable size of a single QR code module is about 0.25mm
This figure depends on the scanner performance one one hand and the printer technology/dpi or display dpi on the other hand. 0.25mm is about the absolute minimum that we can expect to work across common devices.
https://www.qrcode.com/en/howto/cell.html

2. A recommended guideline is module_count = 250 * scan_distance / qrcode_dimension
http://www.growthvine.com/blogs/kelly-flowers/2011/11/28/optimal-size-for-qr-codes-depend-on-a-few-things

If we assume the following:
* QR code size: 35mm. This is the maximum that can be accommodated on a typical mobile device if we consider the required quiet zone. It's also reasonable for a printed QR code.
 * Error correction: L.

Then the maximum capacity of the QR code is:
 * Method 1: version 30 137137) -> 1700 bytes
 * Method 2: version 17 (85*85) -> 504 bytes

Since method 1 is really an upper bound, I would suggest to aim for a payload of ~500 bytes for the happy path scenario where the FHIR bundle follows the recommendations, and allow up to ~1500 bytes for degrade cases, with the understanding that the QR code may be harder to scan.

madaster97:
@jmandel to your point about long-form holder DIDs in QR codes, I'm trying to think about the implications there. 

We would only ever expect holder DIDs to be involved for SIOP flows, so maybe that needs to be taken into account in our compression requirements. Since QR-scanned VCs will never be packaged as VPs, we could probably drop the QR-ready requirement for VCs distributed with a specific holder in mind.

We could replace the following:
> In general, we should aim for all issued VCs to also be compatible with QR-scanning. This way, a wallet app could choose to distribute these VCs via a SIOP workflow OR just presenting the VCs as QR codes.

With:
Issued VCs are only required to be QR-Ready if the VC is issued with no specific holder in mind.  For example:
- If a VC is printed as a QR directly by the issuer, then it will need to be QR-ready.
- A VC is downloaded as a `fhir-backed-vc` from an issuer portal, with no specific holder targeted. 

As a result, QR-ready VCs will never include a Holder's DID (let-alone their long-form DID), since they are required not to be issued for a specific holder.

### Issuer DIDs
On the topic of issuer DIDs, I think we should consider across the board implementing the `.well-known/did-configuration` as the main way of distributing DIDs for issuers. This would mean the following:
- In VCs and VPs, any DID referencing an issuer would be short-form
- The DID Document resolution protocol for issuer DIDs would standardly become a query of the `.well-known/did-configuration` endpoint

This seems like the one part of the QR-Ready VC definition that should just be applied across the board. In general, any entity that needs to verify a signature from an issuer will have access to either:
- The `client_uri` from a SIOP `registration` object when connecting wallet to a lab account
- The `issuerDomain` linked on a VC. We should consider making this a requirement if we go the short-form route.

jmandel:
> Since QR-scanned VCs will never be packaged as VPs, we could probably drop the QR-ready requirement for VCs distributed with a specific holder in mind.

We could define different modes of issuance but (everything else being equal) it'd be nifty if even holder-bound VCs were usable in a print-or-display sharing workflow. So if we can define one set of rules that works in both cases, it'd be a win/win. 

> In general, any entity that needs to verify a signature from an issuer will have access to either...

The functional requirements are "can resolve a DID doc" and "can determine whether the DID is part of a given trust network". Resolution of a DID doc could happen by directly including a long-form DID in a VC; looking up a a short-form DID in  a trust network to resolve the long-form; or following a domain link from a VC to resolve the long-form (.well-known). I think we can ensure these are feasible for VPs (SIOP flow) as well as QR-embedded VCs (with or without an embedded holder -- shouldn't matter for issuer signature validation).

jmandel:
The following rules should make it possible for all VCs to be QR-able, and all VCs to include holder DIDs when known.


#### In SIOP Requests
* include issuer's DID
  * short-form + `#fragment` in `header.kid` 
  * long-form in `.payload.iss`

=> Holders can always resolve the requester's DID doc because the long-form DID is included in the request

#### In SIOP Responses:
* include holder's DID
  * short-form + `#fragment` in `header.kid` 
  * long-form in `.payload.did`

=> Requesters can always resolve the holder's DID doc because the long-form DID is included in the request


#### In Health Card VCs:
* include issuer DID
  * short-form only
  * present in `header.kid` (+ `#fragment`)
  * present in `payload.iss`
* optionally include `payload.issOrigin`
* include `holder.id` if known/linked
  * short-form DID only
* keep size within QR limits by applying all the tricks (deflate, splitting cards, ?maybe protobuf representation)

=>  Verifiers looking at a VC can resolve an issuer's DID Doc as follows:
1. look up the short-form DID in a known trust framework or cache; otherwise
1. look up the short-form DID in the `payload.issOrigin`'s `.well-known/did-configuration.json`; otherwise
1. (future) Look up the short-form DID in a sidetree node; otherwise
1. fail

gillesroux:
> Really nice examples at https://github.com/minvws/nl-eHealth-experimental/tree/master/examples/syntetic-2 using protocol buffers to gain ~2x compression on these kinds of bundles. There's a challenge of agreeing on a "maximal" set of protobuf definitions, because you can't just add in properties later the way you can with JSON. But compelling results re: size!

This is a nice way to save space since we'd be omitting the field names from the output, however it would require to standardize the field name <-> field number mappings. Do you have a suggesting on how to do that? 

A few ideas:
 * Generate field numbers as hashes of field names (however high field numbers are less space-efficient)
 * Rely on the order in which the fields are defined in the json schema at the specified version
 * Specify field numbers as part of the FHIR profile. Is there a way to do that?
 * As last resort, hard-specify the mappings (especially if we decide to only allow a subset of FHIR fields in the QR code)

jmandel:
If we were going to use protocol buffers I assume we would want to use the definitions from https://github.com/google/fhir which are carefully generated and maintained to cover the entire FHIR data model including its extensibility structure. (BTW FHIR does define a canonical field order which is necessary for its XML representation.)

Main challenges I see:
* these definitions cover fhir but we would still have the same problems deciding what fields to allow and how to number them for VC contents outside/around of the fhir bundle -- and in that context we would lose the built-in extensibility of the VC model
* Google's official protobuf structures don't align directly with the fhir json representation (e.g., primitives are kept behind a "value" field) -- victorious with us come up requires a bunch of marshalling/unmarshlling code to put things into FHIR JSON
* working with the full package of profiles can be challenging especially with libraries that try to load definitions dynamically
* this is a major new dependency in the stack and requires implementers to write and test a lot more code, especially those who already have robust support for FHIR JSON
* If we go the protobuf route, I think I would back down on my preference for unifying the health card representation for QRs with the health card representation for other 

Overall I'm wary of introducing a dependency on a schema-dependent binary representation, for these reasons.

madaster97:
@jmandel I agree with the VC rules you have above. This is the only piece I'm concerned with:
> include holder.id if known/linked
> -  short-form DID only

Given that holders and verifiers are exclusively using long-form DIDs for now, this poses a security risk. Since there's nothing verifiable tying a given suffix to a DID document (in a long-form DID), another holder could hijack a VC by just changing it's suffix at presentation time to match the `holder.id` field in the VC.

This restriction is why I think VCs should only need to be "QR-Ready" if there's no holder in mind, since we really can't get away with having short-form holder DIDs securely tying a VC to a holder (at least currently).

jmandel:
@madaster97 

> another holder could hijack a VC by just changing it's suffix at presentation time to match the holder.id field in the VC.

This should not be possible: at presentation, this other holder needs to supply a long-form DID (i.e., a unique suffix plus the data associated with its initial state). When processing a long-form DID, the verifier needs to check that the long-form DID is valid (i.e., that the unique suffix is indeed what you get when you hash the suffix data from the long-form payload) and that it matches the holder's DID (i.e. that the short form matches what's expected). In other words, you can't change your suffix because the long-form DID includes the "suffix data" that (when hashed) determines your suffix.

madaster97:
@jmandel do you think we could get the embedded-QR specific language from the `sans-did` into the spec sooner through this PR? I don't have any issues with what you laid out there.

madaster97:
Only one change I'd suggest, out of all the processing you described I think that deflation is the most controversial. I think we should make deflation optional on issuance, and continue to indicate it with the presence/not of the `zip: deflate` jose header.

jmandel:
Thanks @madaster97. Can you provide a bit more color here: are you suggesting that in the role of an issue where you might have difficulty or prefer not to support DEFLATE compression? Or are you thinking more about whether others might have challenges supporting this feature?

(I ask because without compression an Issuer basically will never be able to fit an entire health card in a single useable QR code. If we don't have mandatory support for some compression scheme, I would be much more inclined to explore different techniques for QR generation -- e.g., a per-IG transformation from fhir bundles into a flattered json structure, or protocol buffers, etc.)

madaster97:
Hey Josh,

So what I mean is that we should keep deflation as our recommended compression scheme, but make that compression optional on issuance. This gives issuers the flexibility to either use card splitting or compression as needed, and maybe both!

It also means that for cards that are small enough for QR without compression, that issuers could just skip compression.

Either way, we're still asking verifiers to support both reading compressed cards and split card sets, I'm just trying to add flexibility in issuing VCs. Since we have the jose header available to convey compression/not, I think this is a fairly innocuous change, and really just simplifies some use cases for issuers.

jmandel:
> what I mean is that we should keep deflation as our recommended compression scheme, but make that compression optional on issuance

I'm following the request, and I want to understand a bit about the rationale behind it. From your own issuer perspective, is compression something you might be concerned about, or is it something you'd be able to support?

If we leave compression as an option for issuers *and* we make it OK to split a Health Card into a Heath Card Set to achieve size limits (without first trying/using compression) I'm concerned this will lead to a worse user experience, and it's something I'd like to prevent.

(I'd be OK with a very narrow range of optionality like: if you can fit full data into a card without compression, then you can skip compression -- but even that makes QRs more data-dense and thus harder to scan, so I'm not exactly sold on it.)

jmandel:
This discussion has been incorporated into #64

PR #55, opened 2021-01-29
Generalize 'Lab' to 'Issuer' to fix #51

### LINKED ISSUE #51: Lab Results vs Immunization Status
Note that the Framework document references a "lab" which seems to cover the COIVID testing use case but not the immunization use case. Perhaps these two user experiences should be separated as they are not exactly the same nor used necessarily in the same way at the same time. This is also not consistent with the primary use case identified in the draft FHIR IG.

jmandel:


docs/index.md
```diff
@@ -571,9 +569,9 @@ Finally, the Health Wallet asks the user if they want to save any/all of the sup
     ```
 
 
-## Presenting Lab Results to a Verifier
+## Presenting Helath Cards to a Verifier
```
p2-apple, line 572:
```suggestion
## Presenting Health Cards to a Verifier
```

docs/index.md
```diff
@@ -676,6 +674,20 @@ The process begins with a QR code or `openid://` link. The only differences are:
...
 ---
 
+# FAQ
+
+## Which clinical data should be considered in decision-making?
+* The data in Health Cards should focus on communicating "immutable clinical facts".
+* Each use case will define specific data profiles.
+  * For COVID-19 Vaccination Credentials, the [SMART Health Cards: Vaccination IG"](http://build.fhir.org/ig/dvci/vaccine-credential-ig/branches/main) defines requirements
+* When Health Cards are used in decision-making, the verifier is responsible for deciding what rules to apply
+  * decision-making rules may change over time as our understanding of the clinical science improves
```
p2-apple, line 684: Shall we add:
* decision-making rules may be subject to rules from international, national and local health authorities

jmandel: Yes, adding this now.

p2-apple (approved):

PR #56, opened 2021-01-29
Document holderDid param on the $HealthWallet.issue operation

jmandel:


docs/index.md
```diff
@@ -298,6 +298,8 @@ By using this URI-based approach, the lab can choose to display a static QR code
     
     This allows the Health Wallet to begin the connection workflow directly, without requiring the user to sign into the lab portal or take any extra steps. This is an optional entry point for the connection workflow; it does not change the subsequent steps.
 
+    Upon successful completion of the OpenID Connect workflow, the Issuer should consider the supplied user DID to be "connected" to the user's account. Issuers MAY allow a single uesr account to be connected to multiple user DIDs (e.g., to support use cases where a user connects multiple health wallets to a single account); alternatively, issuers MAY clear out previously connected DIDs upon the initiation or completion of a new `$HealthWallet.connect` flow.
```
p2-apple, line 301:
```suggestion
    Upon successful completion of the OpenID Connect workflow, the Issuer should consider the supplied user DID to be "connected" to the user's account. Issuers MAY allow a single user account to be connected to multiple user DIDs (e.g., to support use cases where a user connects multiple health wallets to a single account); alternatively, issuers MAY clear out previously connected DIDs upon the initiation or completion of a new `$HealthWallet.connect` flow.
```

p2-apple (commented):
One typo. Did you also want to update CHANGELOG?

jmandel:
> One typo. Did you also want to update CHANGELOG?

Thanks -- committed your typo fix and updated the changelog.

p2-apple (approved):

docs/index.md
```diff
@@ -497,7 +497,7 @@ Finally, the Health Wallet asks the user if they want to save any/all of the sup
     
     The `credentialType` parameter is required.
 
-    The `holderDid` parameter is required. If the supplied value has not previously been connected to the issuer via the `$HealthWallet.connect` API, the issuer responds with a FHIR `OperationOutcome` including the `did-not-connected` code:
+    The `holderDid` parameter is optional. If no value is supplied, the resulting Health Cards are not bound to any specific wallet (i.e., the `.vc.holder` should not be populated). If a value is supplied and this DID has previously been connected to the issuer via the `$HealthWallet.connect` API, then the resulting Health Cards are bound to this holder DID (i.e., the `.vc.holder` should be populated).  If a value is supplied but this DID has not previously been connected to the issuer via the `$HealthWallet.connect` API, the issuer responds with a FHIR `OperationOutcome` including the `did-not-connected` code:
```
jmandel, line 500: @p2 and @madaster97 would like to get your take on this update; we've had requests for workflows that don't force binding health cards to a holder DID, and this seems like a good way to accommodate.

p2-apple: Interesting. This works for me.

madaster97: I think this makes sense, but we may want to add an error-code for `did-bind-required`, so that issuers can programmatically indicate if unbound dids are allowed (or maybe a specific scope, since that seems most appropriate to decide in a SMART app login workflow).

PR #61, opened 2021-02-05
Clarify that VC presentation may or may not be encrypted

dleve123:
The current docs with respect to VC presentation are a bit inconsistent: the
flow diagram suggests that the VCs presented by the Health Wallet to the
Verifier are always encrypted, yet the example SIOP Response Object marked
these VCs "as JWS".

At a high-level, like issuance, I think we want verification to optionally
involve encryption.

Solution

* Update verification docs to specify optional encryption steps
* Mark presented VCs as "JWS or JWE" similar to issuance
* Remove some trailing whitespace


dleve123:
@jmandel I noticed this (I think) issue while updating the docs regarding well-known did configuration. Please let me know if I'm misunderstanding presentation though and feel free to close!

PR #67, opened 2021-02-12
Add QR Prefixes

jmandel:


docs/index.md
```diff
@@ -353,10 +353,16 @@ The following limitations apply when presenting Health Card as QR codes, rather
 
 ### Numerical Encoding
 
-When printing or displaying a Health Card as a QR code, the the JWS string value SHALL be represented as a Numeric Mode value consisting of the characters `0`-`9`. Each character "c" of the JWS is converted into a sequence of two digits as by taking `Ord(c)-45` and treating the result as a two-digit base ten number. For example, `'X'` is encoded as `43`, since `Ord('X')` is `88`, and `88-45` is `43`. (The constant "45" appears here because it is the ordinal value of `-`, the lowest-valued character that can appear in a compact JWS. Subtracting 45 from the ordinal values of valid JWS characters produces a range between 00 and 99, ensuring that each character of the JWS can be represented in exactly two base-10 numeric digits.)
+When printing or displaying a Health Card as a QR code, the the JWS string value SHALL be represented as two segments:
+
+1. An `alphanumeric` mode segment consisting of the fixed string `SHC://`
```
gillesroux, line 358: I assume the intent is for the QR code payload to be a URI. If so, the scheme is case-insensitive (https://tools.ietf.org/html/rfc3986#section-3.1) so it might be worth calling that out so verifier don't get tripped up. I would also prefer using lower-case here since it's more common.

On a related note, what's the purpose of the `\\`? Wouldn't csh:DIGITS be simpler, as is done with `data:` URI?

gillesroux (changes requested):

docs/index.md
p2-apple: Agree, I don't think we need '//'? Also, uppercase or lowercase scheme?

jmandel: On mobile platforms, are apps allowed to register for custom schemes without `://`? Are URIs recognized as routable? My experience was no, but I may be missing something.

jmandel: > I would also prefer using lower-case here since it's more common.

This isn't part of the alphanumeric character set in QR codes, right?

jmandel: > I don't think we need '//'

* iOS: https://developer.apple.com/documentation/xcode/allowing_apps_and_websites_to_link_to_your_content/defining_a_custom_url_scheme_for_your_app indicates that `customscheme:value` URIs should be routable / associable with apps
* https://developer.android.com/guide/topics/manifest/data-element#scheme indicates  the same, though also has the note below:

> Note: Scheme matching in the Android framework is case-sensitive, unlike the RFC. As a result, you should always specify schemes using lowercase letters.

p2-apple: Oooh you're right! You register the scheme alone, but the route is required.

jmandel: I'll make updates per these suggestions.

jmandel: On further discussion, lower case is recommended for Android and using `//` is required for browser redirects (at least) on iOS.

madaster97:
> On further discussion, lower case is recommended for Android and using // is required for browser redirects (at least) on iOS.

I think we should make it explicit that we **don't expect custom URI navigation to work**. It should be an expectation that these QR codes won't be usable by a mobile device's native camera for browser navigations or opening registered app(s).

Trying to rely on `shc://` for browser redirects is going to come with all the same issues that `openid://` had. We shouldn't be asking holders (or anyone hoping to receive health cards) to reside at a predefined URI scheme.

If a mobile app wants to obtain a VC, the appropriate methods are:
1. Having it's own embedded QR scanner
2. Using a file upload/register for extension
3. FHIR API Access

With that, I'm skeptical if we should add this prefix. You can make the argument that verifiers can use it as a "first check" before processing a QR payload, but there are a number of checks they can make early on instead such as:
1. The initial string must be numeric
2. Decoding the numeric (with a pretty simple number -> character mapping) must result in a flattened JWS

We've already heard implementers saying that every byte counts, and I think that adding this prefix doesn't add enough benefit (given assumptions above).

jmandel:
I appreciate the perspective @madaster97 .  Agreed that we shouldn't expect default camera behavior to work, but I don't see why we should prohibit it; if someone wanted to build support into a default camera app, it'd certainly take less custom behavior if an easy, explicit prefix check was available.

@p2-apple how important is this prefix to you?

p2-apple:
I believe _with_ the prefix an OS-level data-detector could at least be considered. Without it one can just guess what that code means, e.g. if the first six digits match x (because they will always be the same for Health Cards) you could _attempt_ our decoding. That won't be usable to build an OS-level data-detector. I'd prefer it if we added the prefix/scheme.

jmandel:
The only downside I see is a ~11 character hit on the JWS size (i.e., ~66 bits of JWS payload data), and this kind of OS level support feels like a useful avenue to keep open.

madaster97:
That makes sense. It's not a big hit to add it, I mostly wanted to make sure we don't try to rely on it for navigation. 

> if someone wanted to build support into a default camera app
This is my lack of mobile knowledge showing... This seems like a valid use case. 

Just to be clear, if no "camera app" is installed on the device, will the OS still try to float an `shc://...` string into a browser? If it won't, then I don't see any issue here.

jmandel:
Yeah, I don't think camera apps generally try to handle an unknown uri scheme by routing them to a system browser -- so I think we're okay.

jmandel:
https://www.nayuki.io/page/optimal-text-segmentation-for-qr-codes is a nifty utility that I used to confirm size limits for v22 QR codes. Merging with  `shc:` prefix.

ChristopherA:
We now have ported Nayuki's code to pure Swift library for iOS/Mac at https://github.com/BlockchainCommons/QRCodeGenerator

There is an iOS demo of the library at https://github.com/BlockchainCommons/QRCodeGeneratorDemo

We have a number of other airgapped QR libraries and specifications at Blockchain Commons, but of interest to this community is we also support multi-frame, aka "animated" QRs with fountain coding for efficient transmission in case of lost frames https://github.com/BlockchainCommons/URKit

There is a video demonstrating this capability at https://www.youtube.com/watch?v=t-GGZ9FyuT8 — a number of cryptocurrency wallets are now using this to perform airgapped signing of bitcoin and other transactions.

PR #68, opened 2021-02-12
a few small grammar and typo fixes

isaacvetter:
no functional changes, just a cleaning following a close read


p2-apple (approved):

jmandel:
Thanks for these! (I think the bullets under "Present" are going to render as numerals but we'll see.)

PR #72, opened 2021-02-24
Added clarification on kid calculation.

christianpaquin:
Explicitly require the base64url-encoded SHA-256 of the thumbprint for the health card `kid`, as this is not mandated by [RFC7638](https://tools.ietf.org/html/rfc7638).


jmandel:
Great catch, thanks!

PR #74, opened 2021-03-01
rejigger multi-part / chunking for QR generation

isaacvetter:


docs/index.md
```diff
@@ -351,17 +338,34 @@ The following limitations apply when presenting Health Card as QR codes, rather
...
 
 (The reason for representing Health Cards using Numeric Mode QRs instead of Binary Mode (Latin-1) QRs is information density: with Numeric Mode, 20% more data can fit in a given QR, vs Binary Mode. This is because the JWS character set conveys only log_2(65) bits per character (~6 bits); binary encoding requires log_2(256) bits per character (8 bits), which means ~2 wasted bits per character.)
 
-When reading a QR code, scanning software can recognize a SMART Health Card from the `shc:/` prefix. Stripping this prefix and decoding the remaining pairs of numerals yields a JWS.
+For example:
+* a single chunk might produce a QR code like `shc:/56762909524320603460292437404460<snipped for brevity>`
```
dleve123, line 365: Thoughts on always including `C` and `N` indicators (chunk metadata), even when the set only contains a single QR code? So

```suggestion
* a single chunk might produce a QR code like `shc:/1/1/56762909524320603460292437404460<snipped for brevity>`
```

While Health Card verifiers could assume the lack of chunk metadata to indicate a set of 1, I think the 4 characters is worth it for the explicitness it provides.

jmandel: I think the one-card case is still designed to be the common case. I don't feel super strongly here but this comes at a cost of 4 characters from the JWS, which can mean be the difference between fitting and not.

isaacvetter: Definitely, the one-QR code case should the the common case! The UX will be so much better with a single code. My $0.02 -- we should optimize for the single QR code scenario, b/c it'll be the most common. This means, leaving out N, C, and the delimiters when there's only a single code.

docs/index.md
```diff
@@ -337,25 +337,35 @@ The following limitations apply when presenting Health Card as QR codes, rather
...
 
-When a Health Card JWS exceeds 1191 characters, it must be split into multiple QR Codes to ensure readability. Each "chunk" of 1191 characters is numerically encoded and prefixed with an ordinal as well as the total number of chunks required to re-assemble the JWS. 
 
-## Multi-part QR codes
+## Creating a QR code (or a set of QR codes) from a Health Card JWS
 
+### Chunking
 
-### Numerical Encoding
+Commonly, Health Cards will fit in a single V22 QR code.  Any JWS longer than 1195 characters is split into "chunks" of length 1191 or smaller; each chunk is then encoded as a separate QR code to ensure ease of scanning.  Each chunk is numerically encoded and prefixed with an ordinal as well as the total number of chunks required to re-assemble the JWS, as described below.
```
isaacvetter, line 345:
```suggestion
Commonly, Health Cards will fit in a single V22 QR code.  Any JWS longer than 1191 characters is split into "chunks" of length 1191 or smaller; each chunk is then encoded as a separate QR code to ensure ease of scanning.  Each chunk is numerically encoded and prefixed with an ordinal as well as the total number of chunks required to re-assemble the JWS, as described below.
```

Josh - this was just a holdover, right?

isaacvetter: Ohhh ... no, this was intentional ... b/c the removal of the chunking data gives us those four characters. Got it.

docs/index.md
dleve123: Sounds good to me – just wanted to surface. sounds like @isaacvetter and team have gotten deeper into the implementation side of splitting/reconstruction, so I trust that the complexity isn't too much and agreed that we should optimize for the single QR code scenario.

PR #78, opened 2021-03-05
add _since on issueVc operation

jmandel:


docs/index.md
```diff
@@ -286,6 +288,23 @@ The `credentialType` parameter is required. By default, the issuer will decide w
 }
 ```
 
+* **`_since`**. By default, the issuer will return health cards of any age. If the Health Wallet wants to request only cards pertaining to data since a specific point in time, it can provide a `_since` parameter with a `valueInstant`. For example, to request only COVID-19 data since March 2021:
```
p2-apple, line 291: Would we want to use a `valueDate` here, because that should be granular enough?

jmandel: Agreed with the point on granularity; I'm mainly just following the pattern from http://hl7.org/fhir/patient-operation-everything.html . I'm happy to go with `valueDateTime` (instead of `valueInstant`) if you prefer since it allow coarser or finer-grained dates (allows a timestamp or a day).

p2-apple: Ah there it is! From a usage perspective I'd think `valueDateTime` would be better (rather than having to append 00:00:00Z all the time if not necessary), however it's also nice to align with what's out there. I'd favor date-time but have no strong opinions.

p2-apple (approved):

docs/index.md
jmandel: OK, updated to valueDateTime

PR #90, opened 2021-03-10
Document vcIndex for issue-vc output to fix #88

### LINKED ISSUE #88: in issue-vc response, resourceLink may not be unique across >1 VC
We could say:

## Option 1 -- use param order
Alternate VCs and the resourceLink values associated with the vc

i.e.

> param: VC1, (Resoure Link for VC1)*, VC2, (Resource Link for VC2)*

downside: it's fragile to depend on order

## Option 2 -- add a property into the Resource Link structure
Or we could add a VC index in the Resource Link structure (making it {vc index, bundledResource, hostedResource)

downside: this also has a dependency on order...

## Option 3 -- re-work the verifiableCredential param
Update verifiableCredential to include two parts: vc and resourceLink

downside: this breaks clients that have a simple way to extract VCs today)

## Other options?

jmandel:


docs/index.md
```diff
@@ -317,14 +317,20 @@ The **response** is a `Parameters` resource that includes one more more `verifia
...
+In the response, an optional repeating `resourceLink` parameter can capture the link between any number of hosted FHIR resources and their derived representations within the verifiable credential's `.credentialSubject.fhirBundle`, allowing the health wallet to explictily understand these correspondences between `bundledResource` and `hostedResource`, without baking details about the hosted endpoint into the signed credential. The optional `vcIndex` value on a `resourceLink` can be used when a response contains more than one VC, to indicate which VC(s) this resource link applies to. The `vcIndex` is a zero-based index of a `verifiableCredential` entry within the top-level `parameter` array.
 
 ```json
 {
   "resourceType": "Parameters",
   "parameter": [{
+    "name": "verifiableCredential",
+    "valueString": "<<Health Cards as JWS>>"
```
p2-apple, line 327:
```suggestion
    "valueString": "<<Health Card as JWS>>"
```

p2-apple (approved):
Looks good, one suggestion (would also apply to the section above).

docs/artifacts/operation-patient-i-health-cards-issue.json
```diff
@@ -49,6 +49,12 @@
     "min": 0,
     "max": "*",
     "part": [{
+      "name": "vcIndex",
+      "use": "out",
+      "min": 0,
+      "max": "*",
```
lcmaas, line 55: There should be only one index per tuple

```suggestion
      "max": "1",
```

jmandel: How come? If a resourceLink applies to more than one card, this is how you'd say that. Or are you suggesting to just repeat the resourceLink for each card? (I'm fine with that.)

p2-apple: Agree that's a good simplification 👍🏼

lcmaas: In the example that is part of this PR, it was listed as a single-valued element, not an array, so I thought this was what you were intending. I have no preference, but if this can be multi-valued then the example will need to be changed instead. I don't really have a preference of one way over the other.

jmandel: Technical point -- in FHIR Parameters, even a 0..* element shows us as separate parameter entries, never an array of ints. i.e. it'd look like:

```
parameter: [{
  name: "x",
  valueInteger: 0
}, {
  name: "x",
  valueInteger: 1
}]
```

... so the example still did match the schema even for 0..*. But I've simplified per your suggestion, and agree with Pascal, this s fine!

lcmaas: Ahh, right!

PR #91, opened 2021-03-11
Add x5c-related topics to keys section

lcmaas:
Per discussion of tying Health Cards into existing X.509 trust frameworks, plus a few related JWK items:

Add x5c as optional JWK parameter.
Add missing EC parameters for JWK.
Clarify private key parameter 'd' must be absent from JWK.
Minor reordering of keys section for clarity.
Add short section on validating certificate binding and trust.
Add short key management concepts section.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -155,9 +155,9 @@ If the Verifier supports PKI-based trust frameworks and the Health Card issuer i
 the Verifier establishes that the issuer is trusted as follows:
 1. Verifier validates the leaf certificate's binding to the Health Card issuer by:
     * matching the `<<iss value from Signed JWT>>` to the value
-of a `uniformResourceIdentifier` entry in the certificate's Subject Alternative Name extension, and
+of a `uniformResourceIdentifier` entry in the certificate's Subject Alternative Name extension
+(see [RFC5280](https://tools.ietf.org/html/rfc5280#section-4.2.1.6)), and
     * verifying the signature in the Health Card using the public key in the certificate.
-(see [RFC5280](https://tools.ietf.org/html/rfc5280#section-4.2.1.6)).
 2. Verifier constructs a valid certificate path of unexpired and unrevoked certificates to one of its trusted anchors
  (see [RFC5280](https://tools.ietf.org/html/rfc5280#section-6)).
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -171,8 +171,8 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
 issuer's JWK set in its jwks.json file. Retired private keys that are no longer used to sign Health Cards SHALL be destroyed.
 Older public key entries that are needed to validate previously
 signed health cards SHALL remain in the JWK set for as long as the corresponding health cards
-are clinically relevant. However, if a private signing key is compromised, then the corresponding public key
-SHALL be removed from the JWK set immediately.
+are clinically relevant. However, if a private signing key is compromised, then the issuer SHALL immediately remove the corresponding public key
+from the JWK set in its jwks.json file and request revocation of all X.509 certificates bound to that public key.
 
 ## Issuer Generates Results
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -122,7 +122,7 @@ Each public key used to verify signatures is represented as a JSON Web Key (see 
 * SHALL have `"crv": "P-256`, and `"x"`, `"y"` equal to the base64url-encoded values for the public Elliptic Curve point coordinates (see [RFC7518](https://tools.ietf.org/html/rfc7518#section-6.2))
 * SHALL NOT have the Elliptic Curve private key parameter `"d"`
 * If the issuer has an X.509 certificate for the public key, SHALL have `"x5c"` equal to an array of one or more base64-encoded (not base64url-encoded) DER representations of the public
-certificate or certificate chain (see [RFC7515](https://tools.ietf.org/html/rfc7515#section-4.1.6)).
+certificate or certificate chain (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-4.7)).
 
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from Signed JWT>>` + `/.well-known/jwks.json`.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -123,7 +123,8 @@ Each public key used to verify signatures is represented as a JSON Web Key (see 
 * SHALL NOT have the Elliptic Curve private key parameter `"d"`
 * If the issuer has an X.509 certificate for the public key, SHALL have `"x5c"` equal to an array of one or more base64-encoded (not base64url-encoded) DER representations of the public
 certificate or certificate chain (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-4.7)).
-If the issuer has more than one certificate for the same public key (e.g. participation in more than one trust community), then a separate JWK entry is used for each certificate with all JWK parameters identical except `x5c`.
+The public key listed in the first certificate in the `"x5c"` array MUST match the public key specified by the `"crv"`, `"x"`, and `"y"` parameters of the same JWK entry.
+If the issuer has more than one certificate for the same public key (e.g. participation in more than one trust community), then a separate JWK entry is used for each certificate with all JWK parameter values identical except `"x5c"`.
 
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from Signed JWT>>` + `/.well-known/jwks.json`.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -127,7 +127,7 @@ If the issuer has more than one certificate for the same public key (e.g. partic
 
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from Signed JWT>>` + `/.well-known/jwks.json`.
 
-The URL at `<<iss value from Signed JWT>>` SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
+The URL at `<<iss value from Signed JWT>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
 
 **Signing keys** in the `.keys[]` array can be identified by `kid` following the requirements above (i.e., by filtering on `kty`, `use`, and `alg`).
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -123,6 +123,7 @@ Each public key used to verify signatures is represented as a JSON Web Key (see 
 * SHALL NOT have the Elliptic Curve private key parameter `"d"`
 * If the issuer has an X.509 certificate for the public key, SHALL have `"x5c"` equal to an array of one or more base64-encoded (not base64url-encoded) DER representations of the public
 certificate or certificate chain (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-4.7)).
+If the issuer has more than one certificate for the same public key (e.g. participation in more than one trust community), then a separate JWK entry is used for each certificate with all JWK parameters identical except `x5c`.
 
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from Signed JWT>>` + `/.well-known/jwks.json`.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -107,23 +107,28 @@ This framework defines a general approach to **representing demographic and clin
 
 ## Generating and resolving cryptographic keys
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -107,23 +107,28 @@ This framework defines a general approach to **representing demographic and clin
...
+* Issuer publish the corresponding public key (public key) at `/.well-known/jwks.json`
+* Wallets and Verifiers use the public key to verify Issuer signatures on Health Cards
 
 ### Determining keys associated with an issuer
+Each public key used to verify signatures is represented as a JSON Web Key (see [RFC 7517](https://tools.ietf.org/html/rfc7517)):
+* SHALL have `"kty": "EC"`, `"use": "sig"`, and `"alg": "ES256"`
+* SHALL have `"kid"` equal to the base64url-encoded SHA-256 JWK Thumbprint of the key (see [RFC7638](https://tools.ietf.org/html/rfc7638))
+* SHALL have `"crv": "P-256`, and `"x"`, `"y"` equal to the base64url-encoded values for the public Elliptic Curve point coordinates (see [RFC7518](https://tools.ietf.org/html/rfc7518#section-6.2))
+* SHALL NOT have the Elliptic Curve private key parameter `"d"`
+* If the issuer has an X.509 certificate for the public key, SHALL have `"x5c"` equal to an array of one or more base64-encoded (not base64url-encoded) DER representations of the public
```
jmandel, line 124: I think https://tools.ietf.org/html/rfc7517#section-4.7 is a more specific reference, since it refers to the use of `x5c` on a JWK structure?

jmandel, line 124: Is it possible for an issuer to belong to >1 trust framework? And if so, is there a standard way to publish >1 chain? Does this matter?

lcmaas: agreed and updated.

lcmaas: Yes, to publish >1 chain you can duplicate the JWK entry with all parameters the same except the x5c. I considered including that, but decided to lean towards 'less is more', at least to start. As far as does this matter, hard to say, but there is a solution.  I'll stick in a short sentence, though, since others may ask the same question...

jmandel: Perfect -- it's a solution whose clunkiness is a good match for the [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it)-ness of the request :)

christianpaquin: Any thoughts about using x5u (URLs) vs. integrating the chain directly in the issuer key set? This could simplify the multi-framework situations. Granted, we would need another cert chain resolution mechanism, but this flexibility might be useful for some applications (e.g., using a trusted key repo)

lcmaas: ok I added in the JWK details section but held off on adding the reciprocal information in the validation section based on above.

docs/index.md
```diff
@@ -142,6 +147,29 @@ The URL at `<<iss value from Signed JWT>>` SHALL NOT include a trailing `/`. For
...
+1. Verifier validates the leaf certificate's binding to the Health Card issuer by:
+    * matching the `<<iss value from Signed JWT>>` to the value
+of a `uniformResourceIdentifier` entry in the certificate's Subject Alternative Name extension
+(see [RFC5280](https://tools.ietf.org/html/rfc5280#section-4.2.1.6)), and
+    * verifying the signature in the Health Card using the public key in the certificate.
+2. Verifier constructs a valid certificate path of unexpired and unrevoked certificates to one of its trusted anchors
+ (see [RFC5280](https://tools.ietf.org/html/rfc5280#section-6)).
+
+
+### Key Management
```
christianpaquin, line 166: Should we add lifetime recommendations? I'd recommend 1 year to match the new norms for endpoint TLS certs.

docs/index.md
```diff
@@ -107,23 +107,28 @@ This framework defines a general approach to **representing demographic and clin
...
+* SHALL have `"kty": "EC"`, `"use": "sig"`, and `"alg": "ES256"`
+* SHALL have `"kid"` equal to the base64url-encoded SHA-256 JWK Thumbprint of the key (see [RFC7638](https://tools.ietf.org/html/rfc7638))
+* SHALL have `"crv": "P-256`, and `"x"`, `"y"` equal to the base64url-encoded values for the public Elliptic Curve point coordinates (see [RFC7518](https://tools.ietf.org/html/rfc7518#section-6.2))
+* SHALL NOT have the Elliptic Curve private key parameter `"d"`
+* If the issuer has an X.509 certificate for the public key, SHALL have `"x5c"` equal to an array of one or more base64-encoded (not base64url-encoded) DER representations of the public
+certificate or certificate chain (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-4.7)).
 
-Issuers SHALL publish keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from Signed JWT>>` + `/.well-known/jwks.json`.
+Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from Signed JWT>>` + `/.well-known/jwks.json`.
```
christianpaquin, line 128: Should we encourage HTTPS here? (SHALL?)

christianpaquin (commented):
I like the approach of tying trust back to a PKI. Simplifies trust management.

docs/index.md
lcmaas: re x5u, I'm not sure there is real benefit over x5c because x5u can still only point one cert or cert chain per the current JWK spec (basically same info as x5c, so no real upside and it requires one additional link dereference.) it's too bad p7b (as allowed in X509 AIA URLs) isn't allowed as an option at x5u endpoint in the JWK and JWS specs. I did add some text in response to an earlier review comment about how multi-framework situations could be handled in a JWKS with multiple JWK entries, though, which hopefully addresses this sufficiently.

docs/index.md
lcmaas: agree. added.

docs/index.md
```diff
@@ -142,6 +148,32 @@ The URL at `<<iss value from Signed JWT>>` SHALL NOT include a trailing `/`. For
...
+2. Verifier constructs a valid certificate path of unexpired and unrevoked certificates to one of its trusted anchors
+ (see [RFC5280](https://tools.ietf.org/html/rfc5280#section-6)).
+
+
+### Key Management
+
+Issuers SHOULD generate new signing keys at least annually. 
+
+When an issuer generates a new key to sign Health Cards, the public key SHALL be added to the
+issuer's JWK set in its jwks.json file. Retired private keys that are no longer used to sign Health Cards SHALL be destroyed.
```
isaacvetter, line 171:
```suggestion
issuer's JWK set in its jwks.json file. Retired private keys that are no longer used to sign Health Cards SHALL be destroyed.
```

Luis, this is problematic. A primary use-case is signing a JWS that's represented as a QR code on paper. Once we remove the public key from the jwks.json, we're invalidating all QR codes signed with that keypair. "no longer used to sign" isn't a sufficient timeframe to ensure that QR codes on paper are no longer in use.

jmandel: I think you may be over-interpreting @isaacvetter  -- the statement in the PR say that *private keys* are destroyed when retired; the public keys remain in the JWKS after key rotation. We can add language to clarify if this is ambiguous.

jmandel: Specifically the PR says:

> Older public key entries that are needed to validate previously signed health cards SHALL remain in the JWK set for as long as the corresponding health cards are clinically relevant. However, if a private signing key is compromised, then the corresponding public key SHALL be removed from the JWK set immediately.

... which I think is clear; when you've had a chance to review, @isaacvetter, see if you agree.

docs/index.md
```diff
@@ -142,6 +148,32 @@ The URL at `<<iss value from Signed JWT>>` SHALL NOT include a trailing `/`. For
 }
 ```
 
+### Certificates
+
+X.509 certificates can be used by issuers to indicate the issuer's participation in a PKI-based trust framework.
+
+If the Verifier supports PKI-based trust frameworks and the Health Card issuer includes the `"x5c"` parameter in matching JWK entries from the `.keys[]` array,
+the Verifier establishes that the issuer is trusted as follows:
+1. Verifier validates the leaf certificate's binding to the Health Card issuer by:
```
dleve123, line 157: Perhaps this a product of my not being as versed in the field, but I find the "leaf" language a bit confusion. Can we reference an RFC if it's a well-known term? If not, can we tweak the language? Are we just referring to the encoded certificate that the `x5c` value?

jmandel: The underlying structure is: a verifier trusts some "trust framework root CA", in the form of a certificate containing the root CA's; this root CA is used to sign some "chain" of intermediate certificates which ends in a cert containing the issuer key, etc:

    Trusted Root Cert --> ["Intermediate" Certs]{0, *} --> ["Leaf" Cert w/ key from  $issuer/.well-known/jwks.json]

So the JWK in the jwks.json file points to a *chain* of certificates from the Leaf up to the Root. The verifier needs to walk up this chain and ensure that signatures are valid... all the way to the top, and also ensure that the root represents a trusted entity, and that the leaf represents the issuer of the SMART Health Card being validated.

There's some background in https://tools.ietf.org/html/rfc5280#section-4.2.1.7 which could be a good reference. Perhaps there are better references?

christianpaquin: > There's some background in https://tools.ietf.org/html/rfc5280#section-4.2.1.7 which could be a good reference. Perhaps there are better references?

Good reference; [section 3.2](https://tools.ietf.org/html/rfc5280#section-3.2) might be a better starting point.

docs/index.md
```diff
@@ -142,6 +148,32 @@ The URL at `<<iss value from Signed JWT>>` SHALL NOT include a trailing `/`. For
...
+    * matching the `<<iss value from Signed JWT>>` to the value
```
jmandel, line 158: Should we ask the verifier to check that the keys in the cert match the "bare key" in the JWK (or, equivalently, ask them to check that the keys in the cert have a thumprint that matches the `kid` from the health card)?

In theory, issuers are violating the JWK spec if these *don't* match; and verifiers who only care about a trust framework shouldn't be able to get in trouble even if there's a mismatch, because the PKI is all internally consistent, no matter what the bare keys contain... so I convinced myself these checks aren't required by verifiers at runtime, but I wanted to raise the question here in case this analysis is missing something.

lcmaas: +1 to your conclusion that a check by the verifier isn't required for the reasons you've stated, though I think I will add a sentence reinforcing that if the issuer includes a cert, it has to match the key in the JWK entry, unless there is an objection.

docs/index.md
dleve123: This is really helpful - thanks for clarifying!

docs/index.md
christianpaquin: Maybe I'm missing something, but having a rogue issuer use non-matching keys yet include a x5c that chains to a trusted root is a great way to issue fake creds. I'd therefore say it's important for verifiers to check the JWK key matches the one in the cert.

docs/index.md
```diff
@@ -142,6 +148,32 @@ The URL at `<<iss value from Signed JWT>>` SHALL NOT include a trailing `/`. For
...
+### Key Management
+
+Issuers SHOULD generate new signing keys at least annually. 
+
+When an issuer generates a new key to sign Health Cards, the public key SHALL be added to the
+issuer's JWK set in its jwks.json file. Retired private keys that are no longer used to sign Health Cards SHALL be destroyed.
+Older public key entries that are needed to validate previously
+signed health cards SHALL remain in the JWK set for as long as the corresponding health cards
+are clinically relevant. However, if a private signing key is compromised, then the corresponding public key
+SHALL be removed from the JWK set immediately.
```
jmandel, line 175: Do we want to say anything about what happens when a JWK is removed from a jwks.json file but a non-revoked cert still exists? I'd like to say issuers should not let this happen (or should try to avoid letting this happen).

jmandel: My mental model here is "bare keys need to work for interop; cert chains provide an optional way to make trust scale."

docs/index.md
p2-apple: I agree with Christian, I think we should call out that while it is REQUIRED that the keys in jwks.json match to the keys in the cert, a verifier is wise to VERIFY that the do.

docs/index.md
lcmaas: Yes, it's worth clarifying that if the keys are compromised, the issuer should also request revocation of any associated certificates. Will add...

docs/index.md
jmandel: > yet include a x5c that chains to a trusted root is a great way to issue fake creds

Walk me through this -- if the keys in the x5c don't match *the keys used in the SMART Health Card*, then the health card validation fails. The cert chain needs to be valid, for sure -- a verifier needs to check the chain of signatures up to the trusted root. The question I wanted to raise in this thread was: what if the keys in the leaf cert don't match the bare keys in the JWK. That's a scenario where the keys in the leaf cert have still been "blessed" by a root CA and have still been used to correctly sign a Health Card, right?

lcmaas: The threat Christian is describing is only possible if the Verifier used the (crv,x,y) parameters in the JWK entry to validate the Health Card signature, and then used the X.509 certificate to validate trust. The PR text requires the Verifier to use the public key from the certificate to validate the signature specifically to prevent this.

lcmaas: To Josh's question, yes, so long as the public key from the certificate is used to verify the signature, such a HC would still be valid and trusted.

jmandel: Okay, so providing a clear and safe algorithm for verifiers is critical; this algorithm can be safe even without checking for bare key :: key-in-cert equivalence, but the details definitely matter!

jmandel:
I'd love to wrap this up for an initial merge this week -- @lcmaas any final  tweaks you want to make? (We can keep working on the details after landing this PR; bet to get the basic scaffolding published.)

lcmaas:
I'm good with merging this now and working on any other details later.

PR #92, opened 2021-03-12
Clarify meaning of >1 credentialType value

lcmaas:
Clarify this is a logical AND, not OR
see https://chat.fhir.org/#narrow/stream/268224-smart.2Finternal-health-cards/topic/CredentialType.20mapping


jmandel:
Thanks for submitting this!

PR #105, opened 2021-03-26
Typos, editorial tweaks, removed ref to an old QR requirement.

christianpaquin:
Capitalized Health Card everywhere, other small edits. Removed reference to a old QR requirement (now that VC can be split in multiple QR codes regardless of original size).


jmandel:
Thanks for these!

PR #108, opened 2021-03-29
Use `nbf`

radamson:
Updates the JWT encoding to use `nbf` instead of `iat` for representing the `issuanceDate` VC property in [accordance with the W3C VC spec](https://w3c.github.io/vc-data-model/#jwt-encoding). Updated the documentation and the generate-examples code for this.

Also updates the [lockfile to version 2](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion) (Newer versions of `npm` use this lockfile version and it was automatically created) and adds a README to the generate-code directory.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.4.2
+
+Replace `iat` with `nbf` in JWT payload encoding
+
 ## 0.4.1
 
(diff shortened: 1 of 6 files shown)


jmandel:
Thanks @radamson! Would you mind also adding a patch entry to the changelog for this? Looks great.

PR #114, opened 2021-04-07
Links to other helpful resources

radamson:
Adds links to a few other helpful resources for specification readers and implementers.  https://github.com/smart-on-fhir/health-cards-tests would be another good candidate to add here.


jmandel:
Thanks! We have a "Demo" link above to c19.cards for now.

PR #122, opened 2021-04-14
Strip 'fixed' payload content to reduce size

jmandel:


dleve123:
I posted [in Zulip](https://chat.fhir.org/#narrow/stream/268224-smart.2Finternal-health-cards/topic/should.20a.20vc.20type.20of.20immunization.20imply.20the.20other.20thre.20types.3F/near/234535202) about this.

**If** this is the change that results in the inflated payload representation not conforming to the W3C VC spec, that I'm 👎  on this. Why?

- There's a first class mechanism in the spec to split a SHC across multiple QRs, so shaving off a few bytes off of the payload seems like less of an urgent problem.
- From a training/technical sales perspective, I think there's value is being to confidently state a SHC **is** a Verifiable Credential (at some point in it's encoding pipeline)

Now maybe there are other changes to the spec that I'm not aware of that already results a SHC being _like a_ VC (but not exactly a VC), in which case, I'm more fine with this.

jmandel:
https://github.com/smart-on-fhir/health-cards/blob/97af15c9eb80800793073ca24049bce9b712e54f/docs/credential-modeling.md#mapping-into-the-w3c-vc-data-model provides more background on the mapping, by making implicit steps explicit. Keep in mind that even if we remove steps 2 and 3 from this process, step 1 means there's still a process. And to provide useful compliance, we'd need to pack the full context from (2), or a URL pointing to that, into every payload as well (something we just haven't tackled yet).

dleve123:
After further discussion, I withdraw my 👎. See this [Zulip message](https://chat.fhir.org/#narrow/stream/268224-smart.2Finternal-health-cards/topic/should.20a.20vc.20type.20of.20immunization.20imply.20the.20other.20thre.20types.3F/near/234542679) for more nuance.

isaacvetter:
my take - this is a good, practical change and saves some space. Of overwhelming priority for us is simply quickly deciding if we should make this change.

jdkizer9:
Confirming that this will have small, if any, impact on CommonHealth.

zsura:
👍

p2-apple (approved):

christianpaquin:
Should the spec version be bumped to 0.4.5?

PR #124, opened 2021-04-16
JWS header should include signing algorithm

isaacvetter:
I think that this is a correction and not a breaking change, right?

[RFC7515](https://tools.ietf.org/html/rfc7515#section-4.1.1)


p2-apple (approved):

jmandel:
Yes indeed -- thanks for catching this!

PR #141, opened 2021-05-07
Harmonized occurrences of "QR codes"

christianpaquin:
Harmonized 'QR code' occurrences, using lowercase c.


jmandel (approved):
Thanks for this!

Issue #82, opened 2021-03-08
Clarify data type for parameter _since


yunwwang:
"By default, the issuer will return health cards of any age. If the Health Wallet wants to request only cards pertaining to data since a specific point in time, it can provide a _since parameter with a valueDateTime (**which is an ISO8601 string at the level of a year, month, day, or specific time of day**)."

The ISO8601 extended time format (eg: yyyy-MM-ddThh:mm:ss) aligns with FHIR dateTime format.
The ISO8601 basic time format (eg: yyyy-MM-ddThhmmss) does not.
Please clarify if which format does the parameter use?

jmandel:
Sure -- it's a FHIR valueDateTime, so it has to be a valid FHIR DateTime. I'll add a link to the FHIR datatype definition and and an example.

PR #142, opened 2021-05-12
Add page for IAL code system

masnick:
This PR adds a page defining a code system for identity assurance level. The page added by this PR will appear at `https://smarthealth.cards/ial` when this PR is merged in. This URI will then be used inside the `fhirBundle` element of the SHC like:

```
...
"resource": {
   "resourceType": "Immunization",
   "meta": {
     "security": [{"system": "https://smarthealth.cards/ial", "code": "IAL1.2"}]
   },
   ...
```

### DIFF
--- a/docs/ial.md
+++ b/docs/ial.md
@@ -1,0 +1,30 @@
+---
+title: "Code System: Identity Assurance Level"
+---
+
+Defining URL
+:    `https://smarthealth.cards/ial`
+
+Version
+:    1.0
+
+License
+: <https://creativecommons.org/licenses/by/4.0/>
+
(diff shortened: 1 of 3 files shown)


jmandel (approved):

christianpaquin:
Should the validation SDK enforce these codes? The page says: `These codes may be used` (should `may` be capitalized?), but should we warn if these codes are not used? (in other words, are these _the_ codes we expect to see?)

jmandel:
For now these requirements are being captured in the Vaccination IG; we're just publishing the codes in a page of the main spec so we have a "landing page" for them.

The idea is that it's fine to omit these codes (in which case you're supposed to omit `.meta` altogether), which implies IAL1 (i.e. implies nothing in particular is known about the identity assurance level). It'd be good for the validator to check these assumptions as part of #82 to start; not sure yet if they should be "universal".

PR #144, opened 2021-05-18
Clarify VC type inclusion.

christianpaquin:
Made the inclusion of the `https://smarthealth.cards#health-card` VC type a MUST, and linked definition page.


docs/index.md
```diff
@@ -216,7 +216,7 @@ Issuer ->> Holder: Holder receives Health Card
...
+The `type`, and `credentialSubject` properties are added to the `vc` claim of the JWT. The `type` values are defined in [Credential Types](https://smarthealth.cards/vocabulary/); the `https://smarthealth.cards#health-card` MUST be present; other ones SHOULD be present as appropriate. The `issuer` property is represented by the registered JWT `iss` claim and the `issuanceDate` property is represented by the registered JWT `nbf` ("not before") claim (encoded as the number of seconds from 1970-01-01T00:00:00Z UTC, as specified by [RFC 7519](https://tools.ietf.org/html/rfc7519)).  Hence, the overall JWS payload matches the following structure (before it is [minified and compressed](#health-cards-are-small)):
```
jmandel, line 219:
```suggestion
The `type`, and `credentialSubject` properties are added to the `vc` claim of the JWT. The `type` values are defined in [Credential Types](https://smarthealth.cards/vocabulary/); the `https://smarthealth.cards#health-card` SHALL be present; other types SHOULD be included when they apply. The `issuer` property is represented by the registered JWT `iss` claim and the `issuanceDate` property is represented by the registered JWT `nbf` ("not before") claim (encoded as the number of seconds from 1970-01-01T00:00:00Z UTC, as specified by [RFC 7519](https://tools.ietf.org/html/rfc7519)).  Hence, the overall JWS payload matches the following structure (before it is [minified and compressed](#health-cards-are-small)):
```

radamson (approved):

PR #176, opened 2021-08-31
Adds TLS requirements for issuer keyset

christianpaquin:
There have been discussions about moving the [proposed TLS requirements](https://github.com/the-commons-project/vci-directory/pull/42) for the VCI directory to the core spec. ~~Here is proposal that explicitly cites the cryptographic requirements (without referring to 3rd party requirements).~~

9/16/2021 update: following some discussions, [bcp195](https://www.rfc-editor.org/info/bcp195) is used as an external requirement source, following usage in, e.g.,  [FHIR R4 security](https://hl7.org/fhir/r4/security.html#http), [QHIN Tech Framework](https://rce.sequoiaproject.org/wp-content/uploads/2021/07/QTF-V1-Draft.pdf). This seems to be a good set of recommendations for the core spec; downstream trust frameworks can always limit the requirements even more if needed. I changed the PR to reflect the language of the QHIN doc, to allow TLS 1.3 (not yet recommended by BCP 195).

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,10 @@
 # Changelog
 
-## 1.0.2 
+## 1.0.3
+
+Updated TLS requirements for issuer key set
+
+## 1.0.2
 
 Updated links to the HL7 Implementation Guide
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.0.3
+## 1.1.0
 
 Updated TLS requirements for issuer key set
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -115,11 +115,11 @@ If the issuer has more than one certificate for the same public key (e.g. partic
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled.
 
 The URL at `<<iss value from JWS>>` SHALL use the `https` scheme, and SHALL use the following TLS version 1.2 or 1.3 configurations:
-* For TLS 1.2, use a ciphersuite conforming to the following algorithm choices and _minimum_ cryptographic strength:
+* For TLS 1.2, support at least one ciphersuite conforming to the following algorithm choices and _minimum_ cryptographic strength:
   * Key exchange: ECDHE using a 256-bit curve
   * Authentication: ECDSA with a 256-bit curve, or RSA with a 2048-bit key
   * Session cipher: AES with a 128-bit key
-  * Cipher mode: CBC or CGM
+  * Cipher mode: CBC or GCM
   * Hash algorithm: SHA256
   * The following recommended ciphersuites adhere to these requirements:
     * TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
--- a/docs/index.md
+++ b/docs/index.md
@@ -113,7 +113,7 @@ The public key listed in the first certificate in the `"x5c"` array SHALL match 
 If the issuer has more than one certificate for the same public key (e.g. participation in more than one trust community), then a separate JWK entry is used for each certificate with all JWK parameter values identical except `"x5c"`.
 
 
-Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled, using TLS version 1.2 following the IETF [BCP 195](https://www.rfc-editor.org/info/bcp195) recommendations or TLS version 1.3.
+Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled, using TLS version 1.2 following the IETF [BCP 195](https://www.rfc-editor.org/info/bcp195) recommendations or TLS version 1.3 (with any configuration).
 
 The URL at `<<iss value from JWS>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -113,9 +113,9 @@ The public key listed in the first certificate in the `"x5c"` array SHALL match 
 If the issuer has more than one certificate for the same public key (e.g. participation in more than one trust community), then a separate JWK entry is used for each certificate with all JWK parameter values identical except `"x5c"`.
 
 
-Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled, and using TLS version 1.2 following the IETF [BCP 195](https://www.rfc-editor.org/info/bcp195) recommendations or TLS version 1.3.
+Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled, using TLS version 1.2 following the IETF [BCP 195](https://www.rfc-editor.org/info/bcp195) recommendations or TLS version 1.3.
 
-The URL at `<<iss value from JWS>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**). The 
+The URL at `<<iss value from JWS>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
 
 **Signing keys** in the `.keys[]` array can be identified by `kid` following the requirements above (i.e., by filtering on `kty`, `use`, and `alg`).
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -112,31 +112,10 @@ certificate or certificate chain (see [RFC7517](https://tools.ietf.org/html/rfc7
 The public key listed in the first certificate in the `"x5c"` array SHALL match the public key specified by the `"crv"`, `"x"`, and `"y"` parameters of the same JWK entry.
 If the issuer has more than one certificate for the same public key (e.g. participation in more than one trust community), then a separate JWK entry is used for each certificate with all JWK parameter values identical except `"x5c"`.
 
-Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled.
 
-The URL at `<<iss value from JWS>>` SHALL use the `https` scheme, and SHALL use the following TLS version 1.2 or 1.3 configurations:
-* For TLS 1.2, support at least one ciphersuite conforming to the following algorithm choices and _minimum_ cryptographic strength:
-  * Key exchange: ECDHE using a 256-bit curve
-  * Authentication: ECDSA with a 256-bit curve, or RSA with a 2048-bit key
-  * Session cipher: AES with a 128-bit key
-  * Cipher mode: CBC or GCM
-  * Hash algorithm: SHA256
-  * The following recommended ciphersuites adhere to these requirements:
-    * TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
-    * TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
-    * TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
-    * TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
-    * TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256
-    * TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384
-    * TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
-    * TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384
-* For TLS 1.3, use one of these ciphersuites:
-    * TLS_AES_256_GCM_SHA384
-    * TLS_AES_128_GCM_SHA256
-* X.509 certificate SHALL use a 256-bit ECDSA or 2048-bit RSA key, must be valid, and trusted by major web browsers.
+Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled, and using TLS version 1.2 following the IETF [BCP 195](https://www.rfc-editor.org/info/bcp195) recommendations or TLS version 1.3.
 
-
-The URL SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
+The URL at `<<iss value from JWS>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**). The 
 
 **Signing keys** in the `.keys[]` array can be identified by `kid` following the requirements above (i.e., by filtering on `kty`, `use`, and `alg`).
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -114,7 +114,29 @@ If the issuer has more than one certificate for the same public key (e.g. partic
 
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled.
 
-The URL at `<<iss value from JWS>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
+The URL at `<<iss value from JWS>>` SHALL use the `https` scheme, and SHALL use the following TLS version 1.2 or 1.3 configurations:
+* For TLS 1.2, use a ciphersuite conforming to the following algorithm choices and _minimum_ cryptographic strength:
+  * Key exchange: ECDHE using a 256-bit curve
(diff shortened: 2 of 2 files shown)


docs/changelog.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 
-## 1.0.2 
+## 1.0.3
```
jmandel, line 3: Given this adds new SHALL-level requirements for issuers, I think it's at least a minor version bump per https://semver.org. Like, issuers who were valid against the 1.0 spec might fail our tests after these changes. (Though, clients who work with 1.0 will all still work after this change, so I can be convinced it's a minor version bump.)

docs/index.md
```diff
@@ -114,7 +114,29 @@ If the issuer has more than one certificate for the same public key (e.g. partic
 
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled.
 
-The URL at `<<iss value from JWS>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
+The URL at `<<iss value from JWS>>` SHALL use the `https` scheme, and SHALL use the following TLS version 1.2 or 1.3 configurations:
+* For TLS 1.2, use a ciphersuite conforming to the following algorithm choices and _minimum_ cryptographic strength:
```
jmandel, line 118:
```suggestion
* For TLS 1.2, support at least one ciphersuite conforming to the following algorithm choices and _minimum_ cryptographic strength:
```

docs/index.md
```diff
@@ -114,7 +114,29 @@ If the issuer has more than one certificate for the same public key (e.g. partic
...
+  * The following recommended ciphersuites adhere to these requirements:
+    * TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
+    * TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
+    * TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+    * TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+    * TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256
+    * TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384
+    * TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
+    * TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384
+* For TLS 1.3, use one of these ciphersuites:
```
jmandel, line 133:
```suggestion
* For TLS 1.3, support at least one of these ciphersuites:
```

docs/index.md
```diff
@@ -114,7 +114,29 @@ If the issuer has more than one certificate for the same public key (e.g. partic
...
+    * TLS_AES_256_GCM_SHA384
+    * TLS_AES_128_GCM_SHA256
+* X.509 certificate SHALL use a 256-bit ECDSA or 2048-bit RSA key, must be valid, and trusted by major web browsers.
```
jmandel, line 136:
```suggestion
* X.509 certificate SHALL use at least a 256-bit ECDSA or at least a 2048-bit RSA key; SHALL be valid; and SHALL be trusted by major web browsers.
```

These are meant as minimums, right? (There's no outer bullet stating such at this point in the bullet list, so we might want to say so explicitly inline.)

jmandel (changes requested):

docs/changelog.md
christianpaquin: I'm ok with that, but this is mostly a deployment requirement: someone could update their TLS config without touching their SHC implementation and (re-)become compliant. The larger question is: what are we asking validation app to do? Reject JWKS if offered at a non-secure enough endpoint? Or is that more the job of, say, a VCI directory auditor to contact issuers to upgrade their deployments.

docs/index.md
christianpaquin: Meaning we're ok if the endpoint supports less secure options, allowing non-conforming clients to still get the keys?

jmandel: I think that's consistent with our internal discussion, yes -- here, I mainly intended it as a disambiguation of the way I already read "use a ciphersuite conforming".

Let me know if you think this is the wrong call.

docs/changelog.md
jmandel: > this is mostly a deployment requirement: someone could update their TLS config without touching their SHC implementation

Yeah, that's fair. I'd say the "implementation" of SMART Health Cards Spec covers quite a bit of the stack, including (as you say) code and deployment configuration. On principle, I think the major semver number on the protocol should reflect spec changes that cause previously-valid implementations to become invalid (whether the required changes mean writing new code, deploying a new version of openssl, setting permissions on a file, etc).

> what are we asking validation app to do? Reject JWKS if offered at a non-secure enough endpoint? Or is that more the job of, say, a VCI directory auditor to contact issuers to upgrade their deployments.

For the validation app, I think we'd log an error if a JWKS if the endpoint doesn't support any of the cyphersuites we require here (i.e., we'd say "the issuer is not compliant with the spec".) Downstream, the VCI directory auditing process would evaluate this (or some heuristic approximating this, like using a well-crafted openssl client config that the VCI directory publishes) when adding new endpoints, and periodically after that.

docs/index.md
```diff
@@ -114,7 +114,29 @@ If the issuer has more than one certificate for the same public key (e.g. partic
 
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled.
 
-The URL at `<<iss value from JWS>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
+The URL at `<<iss value from JWS>>` SHALL use the `https` scheme, and SHALL use the following TLS version 1.2 or 1.3 configurations:
+* For TLS 1.2, use a ciphersuite conforming to the following algorithm choices and _minimum_ cryptographic strength:
+  * Key exchange: ECDHE using a 256-bit curve
+  * Authentication: ECDSA with a 256-bit curve, or RSA with a 2048-bit key
+  * Session cipher: AES with a 128-bit key
+  * Cipher mode: CBC or CGM
```
GaryMcGregor, line 122:
```suggestion
  * Cipher mode: CBC or GCM
```

GaryMcGregor (changes requested):

isaacvetter:
Hey Guys, 

What's the timeline for review before this PR is merged? There's a lot of technical detail. Can it be not before the end of next week?

Isaac

christianpaquin:
> What's the timeline for review before this PR is merged? There's a lot of technical detail. Can it be not before the end of next week?

We should definitely give time for review. Please let us know (anyone) how much time you'd need.

docs/index.md
```diff
@@ -114,7 +114,29 @@ If the issuer has more than one certificate for the same public key (e.g. partic
 
 Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled.
 
-The URL at `<<iss value from JWS>>` SHALL use the `https` scheme and SHALL NOT include a trailing `/`. For example, `https://smarthealth.cards/examples/issuer` is a valid `iss` value (`https://smarthealth.cards/examples/issuer/` is **not**).
+The URL at `<<iss value from JWS>>` SHALL use the `https` scheme, and SHALL use the following TLS version 1.2 or 1.3 configurations:
```
isaacvetter, line 117: I'm uncertain of the intent of this sentence. If a webserver support TLS 1.0, 1.1, 1.2, and 1.3, is that webserver in compliance with this requirement? Could you please clarify?

Generally, are you trying to define a floor or a ceiling? They proposed phrase: "support at least one" instead of "use", below, implies a floor, such that the connection could be insecure if the client accesses the public key doesn't also support these best practices / requirements. 

This PR really begs the question -- why are we putting normative requirements on the server instead of the client?

jmandel: Re: guidance for servers vs clients: if a server doesn't meet these requirements, *nothing* a client does can achieve an appropriate level of protection. So these guidelines are intended to ensure that clients *can* obtain keys in a secure fashion. Clients don't need to retrieve keys directly from an issuer (e.g., they could rely on an intermediary), so I'm not sure it makes sense to impose additional constraints on that side of the interaction.

Re: "ceiling" vs "floor", these terms can be a bit confusing; I *think* the intent of this langauge is to ensure at least one of TLS 1.2 or 1.3 is supported. (Possibly we also want to prohibit servers from supporting TLS 1.0, but that's another question.)

As such, I'd suggest re-wording as:

```suggestion
The URL at `<<iss value from JWS>>` SHALL use the `https` scheme, and SHALL support at least one of the following TLS configurations:
```

docs/index.md
christianpaquin: Thanks, fixed.

lcmaas:
Hi, have we fully considered referring to the best practice documents for TLS that are managed by IETF instead? (BCP 11 and 195); perhaps recommending following these best practices is a good start as this is a nuanced topic to get right, and these docs get updated as deficiencies/vulnerabilities of different cryptographic algorithms are identified. I think this might generally be simpler (and more secure) than trying to list out specific ciphersuites in this guide. To @isaacvetter 's point, I would note that BCP 11 addresses both client and server considerations. 

I will add, however, that using TLS to protect the distribution of an issuer's naked public keys (naked=not in an X.509 certificate) introduces additional security considerations: the strength of the keys and ciphersuites used to protect the TLS connection really need to be at least as strong as the key material being protected, because the TLS connection is being repurposed as a mechanism to securely bind the public key to the identity of the issuer (or in this case to the DNS domain of the issuer). For examples, using a P-256 EC based key agreement protocol or AES-128 encryption for the TLS connection decreases the effective strength of a P-384 key.

I would also note that the TLS connection issues would not be relevant if every JWKS returned included an X509 cert for each public key, because the CA signature on an X.509 cert is already at least as strong as the public key in the cert (for any competent CA) and therefore the keys/certs can actually be safely distributed without TLS at all (because the certificates can be independently validated by the party receiving them). This is commonplace in PKIX communities and also avoids circular trust boot-strapping issues. Unfortunately, as verifier support for certificate based trust of issuer keys is currently optional in the spec, we are probably stuck with explaining that properly configuring and enforcing TLS parameters is critical when either the issuer or verifier relies on TLS to bind public keys to an issuer identity.

christianpaquin:
> BCP 11

christianpaquin:
> Hi, have we fully considered referring to the best practice documents for TLS that are managed by IETF instead? (BCP 11 and 195); perhaps recommending following these best practices is a good start as this is a nuanced topic to get right, and these docs get updated as deficiencies/vulnerabilities of different cryptographic algorithms are identified. 

That would also be a good approach. Better to refer to a standard body's recommendations (such as the [IETF TLS](https://datatracker.ietf.org/doc/draft-ietf-uta-tls-bcp/11/) ones you referred to) than to some vendors (e.g., OpenSSL's, Apple's, like in the VCI PR) best practices, IMO, but these larger recommendations are also very lenient in what they accept, so this PR’s intent was to limit options to a secure min bar (which would need to evolve over time, yes).

> For examples, using a P-256 EC based key agreement protocol or AES-128 encryption for the TLS connection decreases the effective strength of a P-384 key.

Yes, but since SHC mandates a P-256 issuer key, the proposed min bar TLS config matches that security level.

christianpaquin:
It seems we have a decision to make: enforce a MUST support (server must offer the min bar config, but client can ask for less secure options) vs. MUST use (server only offers the min bar config). If we opt for the former, we can adapt the language per @jmandel's comment (or refer to an external document with similar recommendation, per @lcmaas's comment). Comments? Should this be discussed on a call?

docs/index.md
christianpaquin: Given this is a new spec leading to (presumably) new implementations, enforcing a secure min bar makes sense to me. Should be ok for servers, even if keys are hosted by legacy systems. Clients (wallets and newly created apps) should also support this min bar. Bottom line: which legacy components that can't meet these recommendations are we trying to support?

docs/changelog.md
christianpaquin: It will probably be hard for SHC apps and validation tools to check this, because the TLS details are typically handled by  libraries that don't offer visibility of these details.

docs/index.md
jmandel: I think the potential issue is with existing web servers that support TLS 1.1 *in addition to* supporting TLS 1.2 with ciphersuites we recommend. The question is whether that's considered OK or not. Agree it's worth discussing on a call (I expect @isaacvetter may want to join).

isaacvetter:
@christianpaquin , 

> enforce a MUST support (server must offer the min bar config, but client can ask for less secure options) vs. MUST use (server only offers the min bar config).

It's important to note that a new spec != new implementation or deployments. The vast majority of the urls in the directory, for example, are not net-new and also host pre-existing FHIR servers which support existing production integrations supporting patient care. 

Because of this, we'd strongly recommend the "MUST support" option. 

> we can adapt the language per @jmandel's comment (or refer to an external document with similar recommendation, per @lcmaas's comment).

Towards the goal of making it operationally possible to comply with the SHC specification while also scaling support at a multi-national level, it's important that the security requirements be relatively stable. So, if an external document, which? What's the change control process of the external document? The ballot process? The timelines for changes? To avoid answering these questions, it would be nice to adopt Josh's language. 

I'd love a call, if it would be helpful!

Isaac

docs/index.md
```diff
@@ -114,7 +114,29 @@ If the issuer has more than one certificate for the same public key (e.g. partic
...
+    * TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
+    * TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+    * TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
+    * TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256
+    * TLS_ECDHE_ECDSA_WITH_AES_256_CBC_SHA384
+    * TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
+    * TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384
+* For TLS 1.3, use one of these ciphersuites:
+    * TLS_AES_256_GCM_SHA384
+    * TLS_AES_128_GCM_SHA256
```
isaacvetter, line 135:
```suggestion
    * TLS_AES_128_GCM_SHA256
    * TLS_CHACHA20_POLY1305_SHA256
```
Per [rfc8446](https://datatracker.ietf.org/doc/html/rfc8446#section-9.1), why shouldn't `TLS_CHACHA20_POLY1305_SHA256` be in this list?

lcmaas: chacha is not a fips 140-2 approved algorithm so it would likely be a problem for federal entities to require support

christianpaquin: I kept the proposal to the NIST standardized crypto algs, but we can add the CHACHA20_POLY1305.

christianpaquin:
> I'd love a call, if it would be helpful!

I'll send out a scheduling poll to find time next week.

docs/index.md
lcmaas: reword? The way the last phrase of the proposed change is worded suggests that this spec intends to dictate what 'major' web browsers shall or shall not do; also there is likely disagreement on what defines a 'major' browser, so that is not really enforceable/testable.

christianpaquin: Do you have a proposed wording? I was avoiding being explicit for X.509 requirements on purpose here, because this would require _a lot_ of text (what chain properties are valid? Rooting to what? Expiration? Revocation?) Browsers typically have a good handle of what a "valid" cert is; if you use a 'major' one to visit the completed issuer URL and get a green check mark, you are compliant; so could we phrase that in a vendor-neutral way?

christianpaquin: And RE: ECDSA vs EC, an auth X.509 cert can only have ECDSA alg, no? RSA is also a sig alg (confusing in this case, because it's name the same if used for encryption or signing).

christianpaquin:
Seems like Thu Sept 16th 10am Eastern is the best option. I've scheduled a Teams meeting. Let me know if you'd like a calendar invite, otherwise join using [this link](https://teams.microsoft.com/l/meetup-join/19%3ameeting_NWRjNjY2ZWYtNDU3ZC00NTdhLWJkZTgtYzZmNTA4NTZjYTFl%40thread.v2/0?context=%7b%22Tid%22%3a%2272f988bf-86f1-41af-91ab-2d7cd011db47%22%2c%22Oid%22%3a%22a1fdb82b-4807-4c4e-b27a-555e8662e8f4%22%7d).

Questions to discuss (per this PR's threads):
- Should the proposed change bump the spec version to 1.0.3 or 1.1.0 (minor vs. patch number increase)?
- Is the proposed minimum bar ok? In particular, should we only use FIPS(NIST)-certified algs, or are other TLS-valid algs (e.g. CHACHA20_POLY1305) ok?
- Should the minimum bar be required to use, or required to support (offered by the endpoint)?
- Should we specify everything locally in the spec, or offload to another document (e.g., IETF’s recommendation for TLS)

isaacvetter:
Christian,

From some internal conversations, it almost feels like the documented requirements naturally fall into two categories:
1) Pretty rock solid and quite justifiable requirements, e.g. no TLS 1.0 or TLS 1.1. Excluding DES and RCA would fit into this category, for example.
2) More aspirational or recommended best practices, including your list of ciphers. 

I wonder if a reasonable approach would be to require to use/not use #1 and require to support at least some of #2?

Isaac

christianpaquin:
> I wonder if a reasonable approach would be to require to use/not use #1 and require to support at least some of #2?

That indeed sounds like the right approach, we just need to finalize what is in the `#1` and `#2` buckets. After the call next week, I'll update the PR with the emerging consensus, and give a chance to all spec reviewers to take a fresh look.

christianpaquin:
> Seems like Thu Sept 16th 10am Eastern is the best option. I've scheduled a Teams meeting. Let me know if you'd like a calendar invite, otherwise join using [this link](https://teams.microsoft.com/l/meetup-join/19%3ameeting_NWRjNjY2ZWYtNDU3ZC00NTdhLWJkZTgtYzZmNTA4NTZjYTFl%40thread.v2/0?context=%7b%22Tid%22%3a%2272f988bf-86f1-41af-91ab-2d7cd011db47%22%2c%22Oid%22%3a%22a1fdb82b-4807-4c4e-b27a-555e8662e8f4%22%7d).

Just had the call with @isaacvetter, @lcmaas, @GaryMcGregor, @jdkizer9. Seems like there is consensus at referring to an external doc for the details of the TLS requirements, specifically [bcp195](https://www.rfc-editor.org/info/bcp195) (as done, e.g., in the FHIR [security section](http://hl7.org/fhir/R4/safety.html)). I'll give it a detail read, and update the PR.
 
> Questions to discuss (per this PR's threads):
> 
> * Should the proposed change bump the spec version to 1.0.3 or 1.1.0 (minor vs. patch number increase)?

We didn't discuss this much, didn't seem like there was a strong opinion about this. I'd be in favor of only incrementing the patch version, as to not invalidate current implementation.

jmandel:
> We didn't discuss this much, didn't seem like there was a strong opinion about this. I'd be in favor of only incrementing the patch version, as to not invalidate current implementation.

From a compatibility perspective:

* clients get a new documented feature they can rely on(ability to use specific algorithms). This implies a minor version update (at least)
* servers that were previously compatible with our spec may cease being compatible since we're adding new "SHALL" level requirements. This seems to imply a major version update.

I'm having trouble seeing the argument for "patch" only, given these compatibility implications. (To be clear, I don't think this is a huge deal; I just want a consistent approach to the analysis so we know how to apply the same reasoning next time.)

docs/index.md
```diff
@@ -112,7 +112,8 @@ certificate or certificate chain (see [RFC7517](https://tools.ietf.org/html/rfc7
...
 If the issuer has more than one certificate for the same public key (e.g. participation in more than one trust community), then a separate JWK entry is used for each certificate with all JWK parameter values identical except `"x5c"`.
 
-Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled.
+
+Issuers SHALL publish their public keys as JSON Web Key Sets (see [RFC7517](https://tools.ietf.org/html/rfc7517#section-5)), available at `<<iss value from JWS>>` + `/.well-known/jwks.json`, with [Cross-Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin) enabled, using TLS version 1.2 following the IETF [BCP 195](https://www.rfc-editor.org/info/bcp195) recommendations or TLS version 1.3 (with any configuration).
```
jmandel, line 116: Thanks -- "with any configuration" heads off the question I had when I first read this :-)

docs/changelog.md
christianpaquin: Bumped minor version, as suggested

docs/index.md
christianpaquin: Superseded by ref to BCP 195 (TLS 1.1 is now deprecated)

docs/index.md
christianpaquin: Superseded by ref to BCP 195

docs/index.md
christianpaquin: Superseded by ref to BCP 195 (in line with the "at least" suggestion)

docs/index.md
christianpaquin: Superseded by ref to BCP 195.

docs/index.md
christianpaquin: Now allows for any TLS 1.3 config.

jmandel (approved):

PR #179, opened 2021-09-08
Use NIST IAL levels only

aviars:
For this use case, I am strongly recommending using NIST IAL levels ( 1, 2, or 3) and getting rid of "1.x".  This is optional information anyway.  When cases of "low identity assurance", place those as IAL 1.

While the notion of "between 1 and 2" could and should be addressed and it looks like it will be by NIST in revision 4.  See https://github.com/usnistgov/800-63-4/issues/1


jmandel:
See discussion at https://github.com/smart-on-fhir/health-cards/issues/180

PR #187, opened 2021-09-23
Trivial formatting/grammar/shell script fixups from a cursory read-through

gnomon-:
I'm presenting this PR as a set of four commits just in case the corrections I'm offering go against a style guide of which I might be ignorant, to make it easier to cherry-pick specific classes of corrections and ignore others.

Shouldn't be any user-visible changes other than the spelling/grammar fixups in 5a0ddb4.

### DIFF
--- a/docs/credential-modeling.md
+++ b/docs/credential-modeling.md
@@ -48,7 +48,6 @@ According to the procedure above, we start with decisions about FHIR content res
         * Verifiers should not store identity data conveyed via VC, and should delete data as soon as they are no longer needed for verification purposes
         * Verifiers should not expect all elements in the VC to exactly match their own records, but can still use elements conveyed in the VC.
 
-
 ## Mapping into the W3C VC Data Model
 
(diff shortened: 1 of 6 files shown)


docs/credential-modeling.md
```diff
@@ -57,7 +56,7 @@ To create a structure matching the W3C Verifiable Credential [JSON-LD Syntax](ht
 
 2. Add to the `.vc` object:
 
-   ```
+   ```js
```
jmandel, line 59:
```suggestion
   ```
```

I don't think this block is a complete JSON structure (no opening braces) so this format won't quite work

jmandel (changes requested):

docs/credential-modeling.md
gnomon-: Ah good point, thanks.  Fixed up and re-pushed.

jmandel:
Thanks for the close read!

PR #197, opened 2021-10-15
explain how to support Android app intents for app-to-app exchange of SHCs

isaacvetter:
@jdkizer9 & @Opang23

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -249,7 +249,7 @@ To facilitate this workflow, the issuer can include a link to help the user down
 }
 ```
 #### on Android
-When transferring a SMART Health Card from one app to another on Android, use:
+When transferring a SMART Health Card from one app to another on Android, send it via a [content provider](https://developer.android.com/guide/topics/providers/content-providers) URI pointing to the .smart-health-card.
 
 ```java
 Intent intent = new Intent();


docs/index.md
```diff
@@ -248,6 +248,37 @@ To facilitate this workflow, the issuer can include a link to help the user down
   ]
 }
 ```
+#### on Android
+When issuing a SMART Health Card from one app to another on Android, use:
```
jdkizer9, line 252: I'd probably change "issuing" to "transferring" or "sending" to clarify that the sender is not an issuer in the SHC sense.

docs/index.md
```diff
@@ -248,6 +248,37 @@ To facilitate this workflow, the issuer can include a link to help the user down
...
+
+```java
+Intent intent = new Intent();
+intent.setAction(Intent.ACTION_SEND);
+intent.SetType("application/smart-health-card");
```
jdkizer9, line 257:
```suggestion
intent.setType("application/smart-health-card");
```

jdkizer9 (changes requested):
Looks pretty good, left a couple comments

docs/index.md
isaacvetter:
```suggestion
When transferring a SMART Health Card from one app to another on Android, use:
```

docs/index.md
```diff
@@ -248,6 +248,37 @@ To facilitate this workflow, the issuer can include a link to help the user down
...
 }
 ```
+#### on Android
+When transferring a SMART Health Card from one app to another on Android, use:
+
+```java
+Intent intent = new Intent();
+intent.setAction(Intent.ACTION_SEND);
+intent.setType("application/smart-health-card");
+intent.putExtra(Intent.EXTRA_STREAM, uri);
```
jmandel, line 258: What is `uri` here?

docs/index.md
```diff
@@ -248,6 +248,37 @@ To facilitate this workflow, the issuer can include a link to help the user down
...
+intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+startActivityForResult(Intent.createChooser(intent, ...
```
jmandel, line 261: Does this pass a `uri` only, or a complete `.smart-health-card` JSON file?

jmandel: (The `uri` has a limited length and can only convey a single chunk of a single JWS, but the full JSON file can a full JWS and even more than one JWS.)

docs/index.md
```diff
@@ -248,6 +248,37 @@ To facilitate this workflow, the issuer can include a link to help the user down
...
+
+To receive a SMART Health Card in an Android app, register an intent filter, like:
+```xml
+    <intent-filter>
+        <action android:name="android.intent.action.SEND"/>
+        <category android:name="android.intent.category.DEFAULT"/>
+        <data android:mimeType="application/smart-health-card"/>
+    </intent-filter>
+
+<!-- other apps on the device (e.g. Gmail, Files) may use ACTION_VIEW and a binary MIME type, to be safe also support the below -->
```
jmandel, line 272: If apps need to support `ACTION_VIEW` with this MIME type for full compatibility, why don't we just specify this (instead of `ACTION_SEND`) above? Can we describe the advantages of `SEND`?

jmandel: > would also like to support opening a SHC from the Files app & other apps.

If I understand correctly, there are some apps (like GMail) where clicking an SHC attachment file triggers `ACTION_SEND`, and others apps (like Files) where clicking an SHC file in a local directory tree triggers `ACTION_VIEW`, even though these "feel" pretty equivalent to an end-user? If that's the case, it makes sense to recommend that receivers register for both, sure.

docs/index.md
```diff
@@ -248,6 +248,37 @@ To facilitate this workflow, the issuer can include a link to help the user down
   ]
 }
 ```
+#### on Android
+When transferring a SMART Health Card from one app to another on Android, use:
```
isaacvetter, line 252:
```suggestion
When transferring a SMART Health Card from one app to another on Android, implement a [ContentProvider](https://developer.android.com/guide/topics/providers/content-provider-creating) to communicate the uri to a .smart-health-card, like this:
```

isaacvetter: @jdkizer9 , @OpangEpic - I'm unfortunately out of my depth here, gents. Is this content accurate and appropriate?

OpangEpic: I would word it somewhat like this:
> When transferring a SMART Health Card from one app to another on Android, send the SHC via a content provider URI pointing to the .smart-health-card.

If we'd like documentation to point them towards - the content provider overview page may be more apt:(https://developer.android.com/guide/topics/providers/content-providers). They may be using an android-provided content provider (Ex. FileProvider), and not making a custom one.

docs/index.md
jdkizer9: Yes, this explanation sounds right from my experience. `ACTION_SEND` vs `ACTION_VIEW` seems to be more of a "product" decision, so perhaps we can simply identify the differences between the two, include sample implementations, and let implementers choose for themselves. How does that sound to folks?

docs/index.md
isaacvetter:
```suggestion
When transferring a SMART Health Card from one app to another on Android, send it via a [content provider](https://developer.android.com/guide/topics/providers/content-providers) URI pointing to the .smart-health-card.
```

OpangEpic:
@jdkizer9 
From some testing today - I've found a few caveats on the wallet side. I'd like to bounce this off of you James.

1. Some applications seem to send data via intent.setData instead of using the EXTRA_STREAM intent extra. 
I've found that opening an attachment from Gmail is one example. We may want to advise wallet developers to check both areas for a possible uri. 
(Both intent.getData() & intent.getExtras().get(Intent.EXTRA_STREAM)

2. Samsung's default Files app uses an empty MIME type instead of `application/octet-stream` when opening .smart-health-cards or other unknown file extensions. 
This is unfortunate since I think the only way to catch that is with a catch-all `*/*` intent filter. That causes Android to present the wallet app for _all_ file types instead of only ones it can't recognize. That's a big downside if wallets want Samsung Files support. 
The majority of our users are Samsung so I'd expect this to be the primary file app used.

woodynj (approved):

woodynj (approved):

PR #198, opened 2021-10-18
Specified per-SHC revocation

christianpaquin:
We are starting to hear interest in revocation schemes from jurisdictions that are detecting fraud in their own issuance history. In order to promote interoperability, the framework should specify an optional mechanism to enable this.

See related discussion in #204.

~~The proposed way is to add the SHA-256 digest of the card's JWS to a per-`kid` revocation list; this results in a unique string that can be recognized by verifiers. Associating the lists with the corresponding issuing key limits the size of the card revocation list as there will be an upper max limit when the key is rotated.~~

 After further analysis, the approached specified here will not work. Due to [ECDSA malleability](https://yondon.blog/2019/01/01/how-not-to-use-ecdsa/), `hash(JWS)` cannot be used as a revocation identifier, as it can be evaded.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -461,8 +461,6 @@ When reading a QR code, scanning software can recognize a SMART Health Card from
 
 # FAQ
 
-Technical security questions are covered in the [security FAQ page](https://github.com/smart-on-fhir/health-cards/blob/main/FAQ/security.md).
-
 ## Can a SMART Health Card be used as a form of identification?
 
 No. SMART Health Cards are designed for use *alongside* existing forms of identification (e.g., a driver's license in person, or an online ID verification service). A SMART Health Card is a non-forgeable digital artifact analogous to a paper record on official letterhead. Concretely, the problem SMART Health Cards solve is one of provenance: a digitally signed SMART Health Card is a credential that guarantees that a specific issuer generated the record. The duty of verifying that the person presenting a Health Card *is* the subject of the data within the Health Card (or is authorized to act on behalf of this data subject) falls to the person or system receiving and validating a Health Card.
--- a/docs/index.md
+++ b/docs/index.md
@@ -162,22 +162,37 @@ are clinically relevant. However, if a private signing key is compromised, then 
 
 ### Revocation
 
-Individual Health Cards MAY be revoked by an issuer using the digest of their JWS representation. To revoke a card, an issuer:
-1. Calculates the base64url-encoded SHA-256 `<<digest>>` of the utf-8 encoding of the `<<Health Card as JWS>>`
-2. Add the digest to their `https://"<<Issuer URL>>"/.well-known/revoked-shcs-by/kid/"<<kid>>".json`, where
-   - `"<<Issuer URL>>"` is the issuer URL listed in the Health Card,
-   - `"<<kid>>"` is the key ID with which the Health Card was signed,
-   - `"<<kid>>".json" is a JSON file containing an array of revoked Health Cards digests:
-   ```json
-   {
-   "kid": "<<kid>>",
-   "digests": [...,<<digest>>]
-   }
-   ```
+Individual Health Cards MAY be revoked by an issuer using the digest of their JWS representation. The Revocation Digest Fragment (RDF) of a Health Card is calculated as follows:
+1. Calculate the SHA-256 `<<digest>>` of the utf-8 encoding of the `<<Health Card as JWS>>`,
+2. Calculate the base64url encoding of the first 128 bits of resulting `<<digest>>`
 
-Verifiers SHOULD try to download the `https://"<<Issuer URL>>"/.well-known/revoked-shcs-by/kid/"<<kid>>".json` file, and if present, MUST reject the Health Card if its JWS SHA-256 digest is contained in the file.
+To enable per-card revocation, the issuer creates, for each of its keys, a JSON Card Revocation List (CRL) file with the following content:
+```json
+{
+"kid": "<<kid>>",
+"ctr": "<<ctr>>",
+"rdf": [...]
+}
+```
+where
+- `"<<kid>>"` is the ID of the corresponding issuer key,
+- `"<<ctr>>"` is counter indicating how many times this file has been updated; initial value is 1,
+-  `rdf` is an array of revoked cards' digest fragments.
 
-If too many Health Cards have been mistakenly issued under an issuer key, and if individual revocation of such Health Cards is not possible (because the issued JWS were not recorded, or if the affected set is unknown), then an issuer SHOULD revoke its key, and allow users to obtain new Health Cards; limiting the validity period of a key helps to mitigate the adverse effects of this situation.
+To revoke a Health Card issued under the key `"<<kid>>"`, an issuer adds its RDF to the `rfd` array of the corresponding `<<kid>>`'s revocation file. After updating the `rdf` array (with one or more items), the `<<ctr>>` is incremented.
+
+The revocation file is made available at `https://"<<Issuer URL>>"/.well-known/crl/kid/"<<kid>>".json`, where
+- `"<<Issuer URL>>"` is the issuer URL listed in the Health Card,
+- `"<<kid>>"` is the key ID with which the Health Card was signed.
+
+Issuers supporting this revocation method SHALL include in their published JWK set, for each key, a `crl#ctr` encoding the update counter for the corresponding revocation file.
+
+If the `crl#ctr` is present in the Issuer's JWK, verifiers SHALL
+- download the `https://"<<Issuer URL>>"/.well-known/crl/kid/"<<kid>>".json` file (or use a cached version if the counter value has not changed since the last retrieval,
+- calculate its JWS revocation digest fragment as specified above,
+- reject the Health Card if the calculated RDF is contained in the CRL's `rdf` array.
+
+If individual revocation of SMART Health Cards is not possible (because the issued JWS were not recorded, or if the affected set is unknown), then an issuer SHOULD revoke its issuing key, and allow users to obtain new Health Cards; limiting the validity period of a key helps to mitigate the adverse effects of this situation.
 
 ## Issuer Generates Results
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -158,8 +158,26 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
 issuer's JWK set in its `jwks.json` file. Retired private keys that are no longer used to sign Health Cards SHALL be destroyed.
 Older public key entries that are needed to validate previously
 signed Health Cards SHALL remain in the JWK set for as long as the corresponding Health Cards
-are clinically relevant. However, if a private signing key is compromised, then the issuer SHALL immediately remove the corresponding public key
-from the JWK set in its `jwks.json` file and request revocation of all X.509 certificates bound to that public key.
+are clinically relevant. However, if a private signing key is compromised, then the issuer SHALL immediately remove the corresponding public key from the JWK set in its `jwks.json` file and request revocation of all X.509 certificates bound to that public key; verifiers will from then on reject all Health Cards signed using that key.
+
+### Revocation
+
+Individual Health Cards MAY be revoked by an issuer using the digest of their JWS representation. To revoke a card, an issuer:
+1. Calculates the SHA-256 `<<digest>>` of the utf-8 encoding of the `<<Health Card as JWS>>`
+2. Add the digest to their `https://"<<Issuer URL>>"/.well-known/revoked-shcs-by/kid/"<<kid>>".json`, where
+   - `"<<Issuer URL>>"` is the issuer URL listed in the Health Card,
+   - `"<<kid>>"` is the key ID with which the Health Card was signed,
+   - `"<<kid>>".json" is a JSON file containing an array of revoked Health Cards digests:
+   ```json
+   {
+   "kid": "<<kid>>",
+   "digests": [...,<<digest>>]
+   }
+   ```
+
+Verifiers SHOULD try to download the `https://"<<Issuer URL>>"/.well-known/revoked-shcs-by/kid/"<<kid>>".json` file, and if present, MUST reject the Health Card if its JWS SHA-256 digest is contained in the file.
+
+If too many Health Cards have been mistakenly issued under an issuer key, and if individual revocation of such Health Cards is not possible (because the issued JWS were not recorded, or if the affected set is unknown), then an issuer SHOULD revoke its key, and allow users to obtain new Health Cards; limiting the validity period of a key helps to mitigate the adverse effects of this situation.
 
 ## Issuer Generates Results
 
@@ -427,6 +445,8 @@ When reading a QR code, scanning software can recognize a SMART Health Card from
 ---
 
 # FAQ
+
+Technical security questions are covered in the [security FAQ page](https://github.com/smart-on-fhir/health-cards/blob/main/FAQ/security.md).
 
 ## Can a SMART Health Card be used as a form of identification?
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -163,7 +163,7 @@ are clinically relevant. However, if a private signing key is compromised, then 
 ### Revocation
 
 Individual Health Cards MAY be revoked by an issuer using the digest of their JWS representation. To revoke a card, an issuer:
-1. Calculates the SHA-256 `<<digest>>` of the utf-8 encoding of the `<<Health Card as JWS>>`
+1. Calculates the base64url-encoded SHA-256 `<<digest>>` of the utf-8 encoding of the `<<Health Card as JWS>>`
 2. Add the digest to their `https://"<<Issuer URL>>"/.well-known/revoked-shcs-by/kid/"<<kid>>".json`, where
    - `"<<Issuer URL>>"` is the issuer URL listed in the Health Card,
    - `"<<kid>>"` is the key ID with which the Health Card was signed,
--- a/FAQ/security.md
+++ b/FAQ/security.md
@@ -27,6 +27,10 @@ If you don't recognize the key, are they tricking verifiers into thinking you ar
 
 Expired private keys should be deleted, the corresponding public keys should stay in the issuer published key set to allow verifiers to validate health cards issued using them. Revoked private keys (compromised, issued in error, etc.) should be deleted and removed from the published key set.
 
+### Some cards have been erroneously issued, can they be revoked
+
+Individual health cards issued by mistake can be revoked by listing the SHA-256 digest of their JWS is a revocation listed hosted by the issuer.
+
 ## User
 
 ### Can someone steal my health card?
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.2.0
+
+Specified Health Card revocation
+
 ## 1.1.1
 
 Added verifier guidance to ignore unrecognized VC types


docs/index.md
```diff
@@ -158,8 +158,26 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+Individual Health Cards MAY be revoked by an issuer using the digest of their JWS representation. To revoke a card, an issuer:
+1. Calculates the base64url-encoded SHA-256 `<<digest>>` of the utf-8 encoding of the `<<Health Card as JWS>>`
+2. Add the digest to their `https://"<<Issuer URL>>"/.well-known/revoked-shcs-by/kid/"<<kid>>".json`, where
+   - `"<<Issuer URL>>"` is the issuer URL listed in the Health Card,
+   - `"<<kid>>"` is the key ID with which the Health Card was signed,
+   - `"<<kid>>".json" is a JSON file containing an array of revoked Health Cards digests:
+   ```json
+   {
+   "kid": "<<kid>>",
+   "digests": [...,<<digest>>]
```
jmandel, line 174: Thinking about whether we could keep the size of these files manageable women in verifications is large. We could suggest things like taking only the first 128 bits (or fewer bits, or any unique prefix) of a sha256 digest without losing our ability to uniquely identify cards. The client logic would still be simple enough, which is to say as long as you have a fresh revocation file, you can just look for any value in this array that is a prefix of the sha256 of the card you have in hand.

docs/index.md
```diff
@@ -158,8 +158,26 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+   }
+   ```
+
+Verifiers SHOULD try to download the `https://"<<Issuer URL>>"/.well-known/revoked-shcs-by/kid/"<<kid>>".json` file, and if present, MUST reject the Health Card if its JWS SHA-256 digest is contained in the file.
+
+If too many Health Cards have been mistakenly issued under an issuer key, and if individual revocation of such Health Cards is not possible (because the issued JWS were not recorded, or if the affected set is unknown), then an issuer SHOULD revoke its key, and allow users to obtain new Health Cards; limiting the validity period of a key helps to mitigate the adverse effects of this situation.
```
jmandel, line 180:
```suggestion
If individual revocation of SMART Health Cards is not possible (because the issued JWS were not recorded, or if the affected set is unknown), then an issuer SHOULD revoke its issuing key, and allow users to obtain new Health Cards; limiting the validity period of a key helps to mitigate the adverse effects of this situation.
```

christianpaquin:
Dynamic updates would indeed be interesting. One design goal is to keep things simple to deploy for issuers. So far, their identity infrastructure is simple enough (static jwks files), so I’d like to avoid requiring them to setup a live endpoint. A version number or timestamp (surfaced as an etag) could be a good way to version the CRL. We could RECOMMEND only sending the delta CRL if a version/timestamp is requested. X.509 CRLs have a "next update" time, to inform verifiers when to download it next, but that might be too rigid for "emergency" revocation. The presence of a CRL and its version could be advertised in the issuer’s jwks.

As far as size goes, cutting the digest in half to 128-bit (16 bytes) is a no brainer, for security. Cutting it more is possible, but would require more analysis, if we consider adversarial attacks (vs. accidental collisions, which are very unlikely).

pvillela:
> Dynamic updates would indeed be interesting. One design goal is to keep things simple to deploy for issuers ...

I think the following ideas mentioned by @christianpaquin would address my performance concerns related to offline verifier apps:
1. If there is a CRL for a given kid, the issuer MUST advertise the existence of the CRL and its version number by including an optional attribute (e.g., "crlVer") in the corresponding public key object in the jwks.json file.
2. We could RECOMMEND only sending the delta for a kid's CRL if a version number is specified as a query parameter.

In my view, item 1 above is the most important from a verifier's performance perspective.  If an issuer implements only item 1 (and not 2) they can have a 100% static implementation of per-kid revocation lists.

christianpaquin:
I updated this proposal based on comments so far. I added a counter value for the CRL (advertised in an issuer's JWK), which inform verifiers on 1) the revocation support, and 2) the need to re-download.

I haven't addressed the "delta" update yet; I wanted to gauge interest in this functionality  first. If desirable, I was thinking of the following: every time the CRL is updated, the issuer updates the `kid.json` file, but also creates a sibling `kid-ctr.json` file with `ctr` being the counter value, containing only the newly revoked digests.
 
This way, a verifier could either request the full CRL, or only the deltas since the last `ctr` it saw. I was thinking of this approach (vs. query params and etag discussed above) to simplify the issuer implementation (static files only).

jmandel:
We could also consider a scheme using append-only CRL files (e.g., a list of kids separated by newlines) and servers could optionally support https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests (which many static hosting services support out of the box).

The jwks.json could advertise the current file length + hash, and clients could request a range to catch up.

jmandel:
As an example, say the revocation list looks like https://gist.github.com/jmandel/4826077d4e192065199a030e3dc10500

### After first update
  * revocation list looks [like this](https://gist.githubusercontent.com/jmandel/2c489b0f6d64a3966d053e7686d06626/raw/7cb5bf30dd5cf1f332e592851cc8ab96843a1095/revoked-for-kid-3Kfdg-XwP-7gXyywtUfUADwBumDOPKMQx-iELL11W9s.txt)
  * JWKS advertises: `"crl": {"bytes": 64, "hash": "ceb2ff3dc4806c01..."}`
  * Client can 
```
$ curl https://gist.githubusercontent.com/jmandel/4826077d4e192065199a030e3dc10500/raw/b7908dc23321767f268bf20cc03467ce848dde46/revoked-for-kid-3Kfdg-XwP-7gXyywtUfUADwBumDOPKMQx-iELL11W9s.txt \
-H 'Range: bytes=0-64'
28dbefb0554ec0c6d8cfd71b51d5d75866ce9c02afd785e8f16ced034535cd4d
```

### After second update: 
  * revocation list looks [like this](https://gist.githubusercontent.com/jmandel/4826077d4e192065199a030e3dc10500/raw/b7908dc23321767f268bf20cc03467ce848dde46/revoked-for-kid-3Kfdg-XwP-7gXyywtUfUADwBumDOPKMQx-iELL11W9s.txt)
  * JWKS advertises: `"crl": {"bytes": 129, "hash": "860b4abd59930760..."}`
  * Client can 

```
$ curl https://gist.githubusercontent.com/jmandel/4826077d4e192065199a030e3dc10500/raw/b7908dc23321767f268bf20cc03467ce848dde46/revoked-for-kid-3Kfdg-XwP-7gXyywtUfUADwBumDOPKMQx-iELL11W9s.txt \
-H 'Range: bytes=65-'
ff0ebe7bd075b68164509d755ea1ddafcbb82d2dd2c785776001b387bf5b4a3f
```

jmandel:
Note: the story for compressed content encoding together with `bytes` isn't great though. In this scheme, the better format for the append-only revocation files would be binary files where each revoked key is represented by the first 16 bytes of its sha256 hash, so no compression is required/needed and each revoke key spans a (n*16, (n+1)*16) range.

christianpaquin:
Interesting idea. The higher-bit decision we have to make is if we want to make the scheme dynamic or static on the issuer side. If static, then we have a few ideas expressed in this thread to consider. This PR discussion is getting lengthy, so I'll summarize the various options in a stand-alone doc, and share on zulip to hopefully get more engagement from SHC issuers.

jmandel:
Sounds good. We should keep in mind that we don't necessarily need to solve the efficiency problem for all clients in order to establish a system that works efficiently. I.e., i publishers have an easy path to correctness, we can think about downstream services that really optimize the client experience on top of that (for example, an org like VCI interleaving revocation data we consume from multiple issuers and offering an API for downloading diffs  of the whole set in one call -- this could be done dynamically or perhaps even statically with a big old matrix of files).

docs/index.md
```diff
@@ -158,8 +158,41 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
 issuer's JWK set in its `jwks.json` file. Retired private keys that are no longer used to sign Health Cards SHALL be destroyed.
 Older public key entries that are needed to validate previously
 signed Health Cards SHALL remain in the JWK set for as long as the corresponding Health Cards
-are clinically relevant. However, if a private signing key is compromised, then the issuer SHALL immediately remove the corresponding public key
-from the JWK set in its `jwks.json` file and request revocation of all X.509 certificates bound to that public key.
+are clinically relevant. However, if a private signing key is compromised, then the issuer SHALL immediately remove the corresponding public key from the JWK set in its `jwks.json` file and request revocation of all X.509 certificates bound to that public key; verifiers will from then on reject all Health Cards signed using that key.
+
+### Revocation
```
pvillela, line 163: I like this proposed change.  I think it is simple, efficient, and addresses the per-card revocation requirement.

christianpaquin:
> I'll summarize the various options in a stand-alone doc

See #204

docs/index.md
```diff
@@ -158,8 +158,41 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+
+Individual Health Cards MAY be revoked by an issuer using the digest of their JWS representation. The Revocation Digest Fragment (RDF) of a Health Card is calculated as follows:
+1. Calculate the SHA-256 `<<digest>>` of the utf-8 encoding of the `<<Health Card as JWS>>`,
```
pvillela, line 166: An alternative to this would be for the issuer to compute the digest (or its unique prefix) of the JWS **excluding** the signature and to combine that digest with a revocation nbf value, i.e., the revocation identifier would be `<unique prefix of signatureless digest>-<revocation nbf>`. This would revoke, for the given kid, any cards with the given signatureless digest value and also containing an nbf value less than or equal to the revocation nbf.  

As the signature has a random component, issuers are not able to revoke a card using the currently proposed identification scheme unless they keep track of sufficient information for each card issued or if they can recover that information after the fact (e.g., the card is retained by a border agent that suspects fraud).  

This alternative scheme would allow issuers to revoke cards without regard for the card's signature and would also allow a card with the exact same information to be reissued with the same signing key by setting the card's nbf value higher than that of the revocation nbf.

jmandel: For our current profiles, this would leave no guaranteed source of entropy in the inputs for the hash digest. This means it would be feasible to probe (name, birthday, vaccine type, vaccine date) tuples and discover source information for revoked SHCs (especially if you start with a population-level list of actual names and birthdates, which is available from large scale data breaches).

docs/index.md
```diff
@@ -158,8 +158,41 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+{
+"kid": "<<kid>>",
+"ctr": "<<ctr>>",
+"rdf": [...]
+}
+```
+where
+- `"<<kid>>"` is the ID of the corresponding issuer key,
+- `"<<ctr>>"` is counter indicating how many times this file has been updated; initial value is 1,
+-  `rdf` is an array of revoked cards' digest fragments.
```
jfhamme-cccs, line 180: I have a minor complaint regarding the use of RDF because it could be confused with RDF/JSON serialization (https://www.w3.org/TR/rdf-json/).  Furthermore, it is overloaded to refer to a single revocation digest fragment and used as the field name for the array of RDFs.  The latter could be fixed by renaming the field "rdfs", but due to my first point I think a better term could be used.  Perhaps SHC Thumbprint?  Then the CRL field might be "shcts".

docs/index.md
jfhamme-cccs: I'm not sure how feasible it is for someone to correlate large data breaches that may contain names and birthdays with vaccine date and type to perform queries on CRLs.  It is more likely that a fraudster would attempt to query the CRL for their own (or sold) fraudulent SHCs.  But in that scenario, the fraudster would have access to all the non-deterministic data too, and thus able to perform the query in either case. 

I'm more concerned about the complications introduced in this proposal.  In the first variant, the verifier must strip the nbf out of the JSON correctly before computing the digest.  While the second variant avoids this complication with JSON serialization, it relies upon the issuer keeping track of nbs values which, to me, is no different than keeping track of RDFs.  Furthermore, it was mentioned in D204 that the choice to go with a counter versus timestamp in the CRL was due issuer mistakes seen with nbf values, but this proposal relies upon those.

docs/index.md
christianpaquin: Wasn't aware of the w3 RDF, we certainly want to avoid collisions.

docs/index.md
christianpaquin: The idea of having a time interval for invalid usage of a kid is interesting. Could be per SHC before a `nbf` as suggested above, but could be useful for all SHC for a specific time period. E.g., if a large issuer (a state/province) knows that invalid cards have been issued by a rogue insider during a specific period (e.g. one week), then advertising an invalid period would be better than revoking the key itself (therefore, invalidating all issued cards).

christianpaquin:
We are converging on a design. I'd like to have a discussion next week with those interested, to mark this PR as "ready for review". Those who can't make it will have the opportunity to comment on the final proposal before it is merged. Please let me know of your availability here: https://www.when2meet.com/?13543272-5VH8Z

christianpaquin:
Tuesday Nov 9th at 2:30pm Eastern Time is the only availability that everyone can make; let's meet then. Teams [meeting link](
https://teams.microsoft.com/l/meetup-join/19%3ameeting_MzM5ZDIxNDItOTY5MS00OWY3LWJjNTUtMTYzMGJmYTgzOGZk%40thread.v2/0?context=%7b%22Tid%22%3a%2272f988bf-86f1-41af-91ab-2d7cd011db47%22%2c%22Oid%22%3a%22a1fdb82b-4807-4c4e-b27a-555e8662e8f4%22%7d) sent to those with known email addresses; for others (@pvillela), use this link.

pmccabensds:
Hi Everyone.. jumping onto the bandwagon a bit late.  However I wanted to clarify my understanding on what is being proposed.

Can somebody please confirm if the proposed SHA-256 Hash would contain a date related to the issuance, ie. the nbf element.

If this is indeed the case, for Nova Scotia,  we do not currently restrict how many QR codes one can download, meaning that any revocation scheme that involves a hash including an issuance date would not be feasible as we may have several dozen or hundred PVCs with the same data, just a different nbf date.

If I understand the proposal by @pvillela - this would better solve the problem for our use case as we would only have to revoke a single set of data, regardless of the nbf date.

Thanks,
Patrick

pmccabensds:
Thanks Christian.   This is exactly what I'm getting at..
"A newly issued QR code with a different nbf would have in a different digest that would need to be added to the CRL."

In our case, since we could technically have a fraudulent user create 10, 000 QR codes (and we didn't see that as a problem previously)  I would much prefer to be able to revoke all of those using a single hash, instead of 10K hashes just for scalability reasons alone. 

I haven't really dug into the details of the time-based approach to see if that's a better fit.   As mentioned, something along the lines of what @pvillela is suggesting where the hash does not include the NBF date seems like it would address this problem as well.

I'm unfortunately double booked for your meeting this afternoon, would have liked to join to add to the discussion.      However, hoping that our use case can help to influence the direction as I don't think the current proposal can scale, especially where we might already have many thousands of copies already floating around.

jmandel:
>As mentioned, something along the lines of what @pvillela is suggesting where the hash does not include the NBF date seems like it would address this problem as well.

We have not yet identified a way to accomplish this which is both safe (from a privacy perspective) and implementable without some issuance-time tracking of issued artifacts. We will certainly bring this up in our call and can include notes on this point.

bradhead:
> Thanks Christian. This is exactly what I'm getting at.. "A newly issued QR code with a different nbf would have in a different digest that would need to be added to the CRL."
> 
> In our case, since we could technically have a fraudulent user create 10, 000 QR codes (and we didn't see that as a problem previously) I would much prefer to be able to revoke all of those using a single hash, instead of 10K hashes just for scalability reasons alone.
> 
> I haven't really dug into the details of the time-based approach to see if that's a better fit. As mentioned, something along the lines of what @pvillela is suggesting where the hash does not include the NBF date seems like it would address this problem as well.
> 
> I'm unfortunately double booked for your meeting this afternoon, would have liked to join to add to the discussion. However, hoping that our use case can help to influence the direction as I don't think the current proposal can scale, especially where we might already have many thousands of copies already floating around.

In British Columbia, we use the same model... The citizen acquires their SHC self-service and we may issue new nbf, so same concern here, so would be interested in digging in on the time based model.  Our preference in a utopia would be to revoke the individual FHIR immunization record contained in the SHC, but I realize that is very challenging.

christianpaquin:
>Our preference in a utopia would be to revoke the individual FHIR immunization record contained in the SHC, but I realize that is very challenging.

We've been hearing this more and more (many issuers have not recorded their issued QR, so wouldn't be able to revoke existing cards). A time-constrained revocation digest that takes as input the FHIR data might (per @pvillela suggestion) might be the most practical approach.

pmccabensds:
> 
> 
> > As mentioned, something along the lines of what @pvillela is suggesting where the hash does not include the NBF date seems like it would address this problem as well.
> 
> We have not yet identified a way to accomplish this which is both safe (from a privacy perspective) and implementable without some issuance-time tracking of issued artifacts. We will certainly bring this up in our call and can include notes on this point.

Thanks @jmandel 

Regarding the privacy comment.   If I understand correctly, is the concern that one could in theory brute force the their way to knowing that a given set of data was revoked ?  Basically, one could try millions of  combinations of names, dob, vaccine information, etc, create the SHA-256 hash and compare against the list ?

If that understanding is correct,  can you help me understand the actual risk here ?

AndrewSmith247:
Further to the meeting, I am still thinking that signing of the CRL adds value. In an organization like the Province of Nova Scotia, there are many people that have access to the servers hosting our JWKS.json and soon to be CRL but the intersect of those people with parties that have access to the private key is minimal. If a person on a server team wanted to spite someone, they need only hash the FHIR data for a person (captured or built) and inject it into the CRL. They wouldn’t want to know, change or want to change the public key as they want the particular individual to be affected and not people en masse. While I am sure this is an edge case, this sort of targeted attack would be hard to identify but easy to have prevented.

I also feel that we have been making assumptions on the hashing algorithm for the CRL hashes and this information needs to be identified somewhere, perhaps within the CRL. We assume SHA-256 which is the **minimum** acceptable hashing function  for interoperability according to NIST (https://csrc.nist.gov/projects/hash-functions/nist-policy-on-hash-functions) and in the future this may change. Providing the algorithm would also assist other jurisdictions consume our CRL.

christianpaquin:
> many people that have access to the servers hosting our JWKS.json and soon to be CRL

If the host is not secure, then someone could replace the jwks.json file and forge all sorts of SHCs (not with the real private key, but with a fake one corresponding to the "new" public key). Protection of these resources are critical, but I don't think CRLs are more important than jwks.json.

> I also feel that we have been making assumptions on the hashing algorithm for the CRL hashes and this information needs to be identified somewhere, perhaps within the CRL.

The SHC framework already makes hard-coded cryptographic choices; SHA-256 is already used for `kid` calculation and in the ECDSA P256 signature. It is proposed to use the same for CRL.

christianpaquin:
> Tuesday Nov 9th at 2:30pm Eastern Time is the only availability that everyone can make; let's meet then. 

We had a productive meeting with many folks from across the board. In order to inform the next steps and propose an update, we’d like to get more information regarding the capabilities of issuers wrt the proposed options. I’ll create a questionnaire tomorrow and invite issuers to answer it.

christianpaquin:
> In order to inform the next steps and propose an update, we’d like to get more information regarding the capabilities of issuers wrt the proposed options. I’ll create a questionnaire tomorrow and invite issuers to answer it.

Here it is: I'd invite issuers to answer this [questionnaire](https://forms.office.com/r/eGVgThy0Vq).

christianpaquin:
See new revocation proposal in PR #205, superseding this one.

PR #205, opened 2021-11-17
2nd revocation proposal

christianpaquin:
Here is a proposal to add a revocation mechanism to the SHC framework. This PR superseded the proposal from PR #198 (which have been discussed in #204). 

After discussing with the various stakeholders, it is proposed to have a dual approach:
1. A forward-looking approach that is simple and respond to future needs, and
2. A backward-compatible approach that works for currently issued cards.

This PR proposes a forward-looking approach. Discussion #207 proposes a two methods for legacy cards, both compatible with the CRL formats and processing rules by issuer and verifiers (minimizing impact on implementations); these can be discussed and specified separately, and then referenced in the core SHC spec.


docs/index.md
```diff
@@ -162,8 +162,45 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+are clinically relevant. However, if a private signing key is compromised, then the issuer SHALL immediately remove the corresponding public key from the JWK set in its `jwks.json` file and request revocation of all X.509 certificates bound to that public key; verifiers will from then on reject all Health Cards signed using that key.
+
+### Revocation
+
+Individual Health Cards MAY be revoked using a revocation identifier property `rid` encoded in the `vc` claim of the JWT. This should be a short identifier, meaningless to the verifiers; the only constraint is that the identifier SHALL use the base64url alphabet (but doesn’t need to be base64url encoded). Issuers MAY use application-specific user identifiers for this purpose, but since these could be publicly listed in revocation lists, issuers SHOULD use a one-way transformation of the data combined with enough entropy to prevent reversal. It is RECOMMENDED to use the base64url encoding of the first 64 bits of the output of HMAC-SHA-256 (as specified in [RFC 4868](https://tools.ietf.org/html/rfc4868)) using a 256-bit random secret key concatenated with the `<<kid>>`.
```
p2-apple, line 169: Can we impose an absolute upper bound of how long a `rid` can be?

christianpaquin: > Can we impose an absolute upper bound of how long a `rid` can be?

There is a natural goal of keeping these small to 1) still fit in v22 QR, and 2) keep the CRLs small, but we could define an upper bound if we need to enforce this. What would you suggest?

p2-apple: Yeah, agree with the implicit goal. So I presume it only makes sense to have an absolute upper limit. I'm not sure what a good value is, should we aim to allow 256 bits, even though we recommend taking the first 64 only? So, in base64, that's like 43 chars, can we limit to 50 chars? Too much? Too little?

p2-apple: OTOH should we just specify _how_ to calculate the `rid` and be done with it? I'm not sure what the risk of conflict is if we just take the first 64 bits.

christianpaquin: 64 bit is safe to avoid accidental collisions. The only "attacker" in this case would be the issuer itself, who could try to generate collisions, a pointless exercice. Since this does not need to be processed by wallets and verifiers, I don't think we need/should enforce an algorithm; but the recommended hmac approach is a good way to transform an ID meaningful to the issuer into an opaque one to everyone else.

p2-apple: Ok, but if we think 64 bits are enough, which is like ~11 base64 chars, then we can set a much lower limit?

jfhamme-cccs:
Why would the specification allow so much flexibility in the generation of the `rid`s?  I'm concerned that issuers could do it in a poor way, and could result in the privacy concerns expressed earlier.  I realize that the `rid` generation has no bearing on the interoperability of verifiers, but why not require issuers to follow the HMAC-SHA-256 approach?

christianpaquin:
> Why would the specification allow so much flexibility in the generation of the `rid`s? I'm concerned that issuers could do it in a poor way, and could result in the privacy concerns expressed earlier. I realize that the `rid` generation has no bearing on the interoperability of verifiers, but why not require issuers to follow the HMAC-SHA-256 approach?

Different issuers have different capabilities wrt to user profiles, not everyone might have access to a user identifier. A "proxy" issuer might simply sign incoming fhir data from a upstream system, and only remember (log) a timestamp for the SHC, and use that as a `rid`. And yes, because there is not interop issue here, it seems that best thing we can do in the spec is to recommend good behavior.

christianpaquin:
I updated the PR, adding a revocation FAQ with more details.

docs/index.md
```diff
@@ -162,8 +162,45 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+"ctr": "<<ctr>>",
+"rids": [...]
+}
+```
+
+where
+- `"<<kid>>"` is the ID of the corresponding issuer key,
+- `"rid"` identifies the revocation method specified in this framework; legacy cards can use different methods specified in external revocation profiles,
+- `"<<ctr>>"` is counter indicating how many times this file has been updated; initial value is 1,
+-  `rids` is an array of revoked cards' identifiers `rid` (using the base64url alphabet) combined with optional timestamps.
```
jmandel, line 185:
```suggestion
-  `rids` is an array of revoked cards' identifiers `rid` values. These values are represented as strings from the base64url alphabet, plus an optional timestamp suffix consisting of `.` followed by a numerical timestamp (e.g., `.1636977600`)
```

docs/index.md
```diff
@@ -162,8 +162,45 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+
+As an example, the `rids` array `[AQPCj4wwk6Mt, lHKzqFUMjhs.1636977600]` marks as revoked any Health Cards with `rid` equal to `AQPCj4wwk6Mt` and Health Cards with `rid` equal to `lHKzqFUMjhs` issued before November 15, 2021 12:00:00 PM GMT.
+
+The per-key revocation file is made available at `https://"<<Issuer URL>>"/.well-known/crl/kid/"<<kid>>".json`, where
+- `"<<Issuer URL>>"` is the issuer URL listed in the Health Card,
+- `"<<kid>>"` is the key ID with which the Health Card was signed.
+
+Issuers supporting this revocation method SHALL include in their published JWK set, for each key, a `crl-version` field encoding the update counter "<<ctr>>" for the corresponding revocation file.
+
+If the `crl-version` is present in the Issuer's JWK for key `<<kid>>`, Verifiers SHALL
```
jmandel, line 197: Can we call this `crlVersion`? Would like to follow a common JSON naming schemes compatible with JS variable names...

docs/index.md
```diff
@@ -162,8 +162,45 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+- Download the `https://"<<Issuer URL>>"/.well-known/crl/kid/"<<kid>>".json` file or use a cached version if the counter value has not changed since the last retrieval,
+- Reject the Health Card if the calculated `rid` is contained in the CRL's `rids` array and the Health Card’s `nbf` is value is before the timestamp, if present.
```
jmandel, line 199:
```suggestion
- Reject the Health Card if the calculated `rid` is contained in the CRL's `rids` array and (if a timestamp suffix is present) the Health Card’s `nbf` is value is before the timestamp.
```

jmandel (changes requested):

FAQ/security.md
```diff
@@ -27,6 +27,10 @@ If you don't recognize the key, are they tricking verifiers into thinking you ar
 
 Expired private keys should be deleted, the corresponding public keys should stay in the issuer published key set to allow verifiers to validate health cards issued using them. Revoked private keys (compromised, issued in error, etc.) should be deleted and removed from the published key set.
```
pvillela, line 28: Can we add an optional `exp` field to the key object in the `jwks.json` to allow for a key that is known to have been compromised after a certain date?  In that case, only SHCs signed with that key after the `exp` value would be automatically invalidated.

FAQ/revocation.md
```diff
@@ -0,0 +1,38 @@
...
+
+Issuers that keep track of every single issued SHC could create a per-SHC `rid` for fine-grained revocation. In many cases, an issuer will have an internal user ID that can be used to revoke all cards belonging to a particular user; using the timestamp feature allows an issue to invalidate cards up to a certain time.
+
+### Why is a one-way transformation on the user ID recommended for revocation ID?
+
+Publishing an internal user ID might be a privacy issue. A one-way transform with high-entropy input prevents reversal of the CRL’s content. The proposed HMAC-SHA-256 algorithm using a 256-bit key achieves that.
+
+### Why are Card Revocation Lists tied to a key identifier?
+
+Since SHC don’t have expiry dates, public keys and revocation information must be publicly available forever. Creating a per-kid CRL allows issuers to cap the size of CRLs, and verifier apps might not need to download the CRLs of old keys when the corresponding SHCs are replaced by newer ones.
```
pvillela, line 17: Can we allow an optional `exp` field to designate an expiration date for an SHC?  This may not be needed in general but it can be very useful to enable a country to issue temporary vaccination certificates to foreign visitors.

baptistelaget: The `exp` field is already usable.

Per the existing [SHC Modeling spec](https://spec.smarthealth.cards/credential-modeling/#mapping-into-the-w3c-vc-data-model):
> Process the payload according to [JWT Decoding Rules](https://www.w3.org/TR/vc-data-model/#jwt-decoding)

Le linked page specifies `exp` and `expirationDate` that matches that use case.

FAQ/security.md
christianpaquin: Putting an expiration date on a key has been discussed, and could be useful on its own merit, but I'd tackle that in a separate thread (PR). I do think that limiting the time range of a key for revocation purposes would be useful. True @baptistelaget, a key leak would allow back signing, but an unauthorized key access (e.g., by a rogue employee having access to an issuance system for a limited time period, without access to the private key itself) could be identified up to a certain datetime, and the adverse effects of the revocation could be limited by keeping a subset of the issued SHCs valid. This being said, I'd rather push this PR as is, and debate the need for this feature separately; short key lifetimes might be enough to address this use case without adding complexity to the revocation system.

FAQ/revocation.md
christianpaquin: Valid question, but should be discussed separately. You might want to open an issue on the spec to discuss @pvillela. Even if the VC spec supports it, you won't see validation apps processing this field unless the SHC explains it.

FAQ/security.md
pvillela: Good points, @baptistelaget and @christianpaquin.  Agree with pushing the PR as-is with respect to this.

FAQ/revocation.md
pvillela: @christianpaquin, can you send me a link to where I can open the issue?

christianpaquin: You can open an issue right on this github repository. See also this [zulip](https://chat.fhir.org/#narrow/stream/284830-smart.2Fhealth-cards/topic/SHC.20Expiration.20Claim) thread discussing SHC expiration.

docs/index.md
christianpaquin: Done.

generate-examples/src/index.ts
```diff
@@ -20,6 +19,10 @@ interface BundleInfo {
   types: string[];
 }
 
+// revocation data
+const issuerSupportingRevocation = new Set([0]); // only one issuer (kid) supports it
+const revocationHmacSecretKey = crypto.randomBytes(32); // 256-bit HMAC key for calculating revocation IDs
```
jmandel, line 24: Should we grab this from config the way we grab private keys (e.g., https://github.com/smart-on-fhir/health-cards/blob/main/generate-examples/src/config/issuer.jwks.private.json)? Otherwise won't the `rid` in our example SHCs change each time the generation script runs?

christianpaquin: Yes, `rid` values would change each time, but they would be updated everywhere, and wouldn't affect card processing. We can certainly pre-generate the secret to keep these values static.

jmandel: Yeah, I think using a static HMAC secret would give us a way to publish examples of revoked cards.

christianpaquin: My point was that when you publish the revoked examples and CRLs, they would have consistent rid values even if different from run to run, because the script re-generate all of these artefacts at the same time. Moot point, I updated the code to use a static secret.

christianpaquin:
I've updated the example generation code to create a revoked card and a CRL. See the updated artefacts on this PR for a better illustration of the proposed feature in action.

docs/index.md
```diff
@@ -162,8 +162,45 @@ When an issuer generates a new key to sign Health Cards, the public key SHALL be
...
+- `"<<ctr>>"` is a counter indicating how many times this file has been updated; initial value is 1,
+-  `rids` is an array of revoked cards' identifiers `rid` values. These values are represented as strings from the base64url alphabet, plus an optional timestamp suffix consisting of `.` followed by a numerical timestamp (e.g., `.1636977600`)
+
+To revoke a Health Card issued under the key `"<<kid>>"`, an issuer adds its revocation identifier to the `rids` array of the corresponding `<<kid>>`'s revocation file. Since an issuer might want to invalidate a series of Health Cards associated with the user up to a certain time, the `rid` might be followed by a separator `.` a timestamp (encoded as the number of seconds from 1970-01-01T00:00:00Z UTC, as specified by RFC 7519). After updating the `rids` array (with one or more items), the `<<ctr>>` is incremented.
+
+As an example, the `rids` array `[AQPCj4wwk6Mt, lHKzqFUMjhs.1636977600]` marks as revoked any Health Cards with `rid` equal to `AQPCj4wwk6Mt` and Health Cards with `rid` equal to `lHKzqFUMjhs` issued before November 15, 2021 12:00:00 PM GMT.
```
jmandel, line 189:
```suggestion
As an example, the `rids` array `["AQPCj4wwk6Mt", "lHKzqFUMjhs.1636977600"]` marks as revoked any Health Cards with `rid` equal to `AQPCj4wwk6Mt` and Health Cards with `rid` equal to `lHKzqFUMjhs` issued before November 15, 2021 12:00:00 PM GMT.
```

jmandel (approved):

PR #214, opened 2021-12-03
Expiration of Health Cards

pvillela:
Introducing the ability to expire Health Cards and the rationale for doing so in special situations.


docs/index.md
```diff
@@ -428,6 +428,14 @@ For example:
...
+Following is a use case for issuing Health Cards that have an expiration date.  Country X needs to issue Health Card vaccination certificates (visitor vaccination passes) to foreign visitors for their use while in the country.  The visitor vaccination certificate is issued based on document(s) from the visitor's country of residence presented by the visitor to Country X.  While there may be reasonable protocols for verification of the authenticity of the source information, Country X may not want to vouch for the validity of the information on the visitor pass for an unlimited period of time.  For example, if at some point after the visitor has left Country X, there is reason for questioning or revoking the document that was used as the source for the visitor pass, then it could be very hard for Country X to learn about the problem and/or revoke the visitor pass.  As a result, a visitor pass based on questionable information would continue to be valid.  This risk can be addressed/mitigated if Country X can set an expiration date on the visitor pass at the time of issuance.  The expiration date would, in this case, correspond to the visitor's allowed duration of stay in Country X.
```
p2-apple, line 435:
```suggestion
One use case to issuing Health Cards with an expiration date is a government entity issuing a vaccination Health Card to foreign visitors for their use while in the country. This visitor's vaccination card is issued based on original documents presented by the visitor. Even with robust verification protocols, the government entity may not want to vouch for the validity of the visitor pass for an unlimited period of time. Importantly, the original document may be invalidated at some point in the future, e.g. by its signing keys being revoked. It may be impractical for the government entity issuing the visitor pass to track and reactively revoke the visitor pass. This risk can be mitigated by setting an expiration date on the visitor pass at the time of issuance. The expiration date could for example correspond to the visitor's allowed duration of stay in the foreign country.
```

docs/index.md
```diff
@@ -428,6 +428,14 @@ For example:
...
+
+To address use cases such as the preceding one, an optional Health Card expiration date can be represented by the registered JWT `exp` claim (encoded as the number of seconds from 1970-01-01T00:00:00Z UTC, as specified by [RFC 7519](https://tools.ietf.org/html/rfc7519)).  Verifiers are required to check the expiration, if present, and reject Health Cards with an `exp` value that is before the current verification date-time.
```
p2-apple, line 437:
```suggestion
To address use cases such as the preceding one, an optional Health Card expiration date can be represented by the registered JWT `exp` claim (encoded as the number of seconds from 1970-01-01T00:00:00Z UTC, as specified by [RFC 7519](https://tools.ietf.org/html/rfc7519)). Verifiers are required to check the expiration, if present, and reject Health Cards with an `exp` value that is before the current verification date-time.
```

p2-apple (commented):
Adding some suggestions to re-wording.

docs/index.md
```diff
@@ -428,6 +428,14 @@ For example:
 
 When reading a QR code, scanning software can recognize a SMART Health Card from the `shc:/` prefix. Stripping this prefix and the following `<ordinal>/<count>/` and decoding the remaining pairs of numerals yields a JWS.
 
+## Expiration of Health Cards
+
+Normally, Health Cards do not expire.  Health Cards are usually intended to contain factual information that is assured to be correct at the point of issuance and should not change with the passage of time.  There are, however, situations where the ability to set an expiration date is beneficial.  To be clear, these are special situations and expirations are **NOT** expected to be routinely used when clinical data Health Cards are issued by an authoritative source.
```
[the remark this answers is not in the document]
pvillela: @jmandel requested the inclusion of the note about expirations not being routinely used.

p2-apple, line 433: Ah ok, then I would suggest adding something like:

```suggestion
SMART Health Cards contain factual information that is assured to be correct at the point of issuance and does not change with the passage of time. Therefore, **Health Cards generally do not expire** and an expiration date is not used. There are, however, situations where the ability to set an expiration date is beneficial.
```

pvillela: I'm good with this rephrasing proposed by @p2-apple if @jmandel is OK with it.

docs/index.md
```diff
@@ -428,6 +428,14 @@ For example:
...
+
+One use case to issuing Health Cards with an expiration date is a government entity issuing a vaccination Health Card to foreign visitors for their use while in the country. This visitor's vaccination card is issued based on original documents presented by the visitor. Even with robust verification protocols, the government entity may not want to vouch for the validity of the visitor pass for an unlimited period of time. Importantly, the original document may be invalidated at some point in the future, e.g. by its signing keys being revoked. It may be impractical for the government entity issuing the visitor pass to track and reactively revoke the visitor pass. This risk can be mitigated by setting an expiration date on the visitor pass at the time of issuance. The expiration date could for example correspond to the visitor's allowed duration of stay in the foreign country.
```
jmandel, line 435:
```suggestion
One use case for issuing SMART Health Cards with an expiration date is a government entity issuing a vaccination card to foreign visitors for their use while in the destination country. This visitor's vaccination card is issued based on original documents presented by the visitor. Even with robust verification protocols, the government entity may not want to vouch for the validity of the visitor pass for an unlimited period of time. Importantly, the original document may be invalidated at some point in the future, e.g. by its signing keys being revoked. It may be impractical for the government entity issuing the visitor pass to track and reactively revoke the visitor pass. This risk can be mitigated by setting an expiration date on the visitor pass at the time of issuance. The expiration date could, for example, correspond to the visitor's allowed duration of stay in the foreign country.
```

docs/index.md
```diff
@@ -428,6 +428,14 @@ For example:
...
+
+To address use cases such as the preceding one, an optional Health Card expiration date can be represented by the registered JWT `exp` claim (encoded as the number of seconds from 1970-01-01T00:00:00Z UTC, as specified by [RFC 7519](https://tools.ietf.org/html/rfc7519)). Verifiers are required to check the expiration, if present, and reject Health Cards with an `exp` value that is before the current verification date-time.
```
jmandel, line 437:
```suggestion
To address use cases such as the preceding one, an optional SMART Health Card expiration date can be represented by the registered JWT `exp` claim (encoded as the number of seconds from 1970-01-01T00:00:00Z UTC, as specified by [RFC 7519](https://tools.ietf.org/html/rfc7519)). Verifiers are required to check the expiration, if present, and reject SMART Health Cards with an `exp` value that is before the current verification date-time.
```

jmandel:
Thanks! I made a couple of tiny edits -- very happy with how this has shaped up.

p2-apple:
> This impracticality applies generally to any large Issuer of health card data and is relevant to: #205

Sure. In this case it's compounded by the fact that the issuer has to monitor upstream certificate revocations (or whatever the upstream revocation story is) and then perform the downstream revocations.

jmandel:
Makes sense to me @christianpaquin  re: example! I see this PR is tied to a branch in @pvillela so I'm not sure whether you have permission to make the change there (if not, we can handle it in a separate PR).

Will plan to merge this one tomorrow unless anyone raises a concern.

christianpaquin:
The `pvillela:main` branch isn't up to date with the current head of the repository, so I can't add the `exp` to the new example-03 series. I don't think I have write access anyway, so I can do this on a separate PR.

pvillela:
@christianpaquin, my fork's main is now up-to-date with your repo.

christianpaquin:
I just added expiration value to the example-03 card.

jmandel (approved):

PR #218, opened 2021-12-08
Clarified revocation HMAC secret use.

christianpaquin:
Clarified that HMAC secret is generated once and reused across keys and SHCs, as suggested by @pvillela [here](https://github.com/smart-on-fhir/health-cards/commit/05984728f4cd553ae0304f36669ed730cb532be8#r61560696).


christianpaquin:
Moved content to PR #230.

PR #224, opened 2022-01-14
Create 003-vaccine-exemptions.md

jmandel:


jpp9:
looks great. one suggestion:

the doc conveys one of the challenges with exemptions as you read through (the intent of SHCs is to convey clinical facts, not a ‘status’ like is a person exemption from vaccination, but in this case the clinical facts themselves are in conflict with the privacy objectives), but might be good to spell that out in the warning, eg:

*Conveying "exemption status" is beyond the SMART Health Cards design intent.** On the other hand, conveying the clinical (or other relevant) facts behind an exemption raises significant privacy concern. If jurisdictions intend to move ahead in this space, VCI has developed a set of policy and technical considerations.

and a couple of small fixes:

+ _In real-world travel use cases, the answer has been "no"_  --> change to _predominantly "no"_  as there are in fact exemptions, but they are quite rare


+ _*Note: It doesn't help use an "unpublished data dictionary"_  —>typo, should be _It doesn't help to use_

jmandel:
Thanks for the close read! I incorporated your suggestions and will merge as a draft RFC.

PR #235, opened 2022-09-11
Updates to reflect vocabulary move to terminology.smarthealth.cards

masnick:


jmandel (approved):
Thanks for the documents and bug fixes along with this redirect!

PR #239, opened 2022-12-06
Clarify scope including patient summaries; deprecate chunking

jmandel:
Based on discussion in [#IPA](https://chat.fhir.org/#narrow/stream/261969-IPA), I'm proposing some updates to the SHC spec to clarify that representing patient summary documents is an intended use case. This PR aims to:

* Provide more recent (2022-era) context and background on the spec
* Ensure that any FHIR bundle type can be embedded in an SHC
* Deprecate "QR chunking" protocol, which has not seen wide adoption in real products and does not support a clean UX for presentation

### DIFF
--- a/docs/credential-modeling.md
+++ b/docs/credential-modeling.md
@@ -13,7 +13,9 @@ Once we make these decisions, we can construct a VC with a **credential subject*
 
 * `credentialSubject` with these top level elements:
     * `fhirVersion`: a string representation of the semantic FHIR version the content is represented in (e.g. `1.0.*` for DSTU2, `4.0.*` for R4, where `*` is a number, not a literal asterisk)
-    * `fhirBundle`: a FHIR `Bundle` resource that includes all required FHIR resources (content + identity resources)
+    * `fhirBundle`: a FHIR `Bundle` resource that includes all required FHIR resources (content + identity resources). For the [`Bundle.type`](http://hl7.org/fhir/bundle-definitions.html#Bundle.type), implementers should choose `collection` unless a more specific type applies (e.g. `document`).
+
+
 
 Resulting payload for the `"credentialSubject"`:
 
--- a/docs/credential-modeling.md
+++ b/docs/credential-modeling.md
@@ -13,7 +13,7 @@ Once we make these decisions, we can construct a VC with a **credential subject*
 
 * `credentialSubject` with these top level elements:
     * `fhirVersion`: a string representation of the semantic FHIR version the content is represented in (e.g. `1.0.*` for DSTU2, `4.0.*` for R4, where `*` is a number, not a literal asterisk)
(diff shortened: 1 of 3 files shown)


docs/credential-modeling.md
```diff
@@ -13,7 +13,7 @@ Once we make these decisions, we can construct a VC with a **credential subject*
 
 * `credentialSubject` with these top level elements:
     * `fhirVersion`: a string representation of the semantic FHIR version the content is represented in (e.g. `1.0.*` for DSTU2, `4.0.*` for R4, where `*` is a number, not a literal asterisk)
-    * `fhirBundle`: a FHIR `Bundle` resource of type "collection" that includes all required FHIR resources (content + identity resources)
```
p2-apple, line 16: Do we not care about `type` at all or do we have suggestions on what to use?

docs/index.md
```diff
@@ -271,7 +273,7 @@ Issuers SHALL ensure that the following constraints apply at the time of issuanc
     * payload is minified (i.e., all optional whitespace is stripped)
     * payload is compressed with the DEFLATE (see [RFC1951](https://www.ietf.org/rfc/rfc1951.txt)) algorithm before being signed (note, this should be "raw" DEFLATE compression, omitting any zlib or gz headers)
    
-For Health Cards that will be exchanged as QR codes, issuers SHALL ensure that:
+For Health Cards that will be directly embedded in QR codes, issuers SHALL ensure that:
```
p2-apple, line 276: Maybe "represented as"?

p2-apple (approved):

docs/credential-modeling.md
jmandel: We want to allow for use cases that cover pure collections as well as document bundles. We could limit to those two, or recommend those two... But I don't necessarily see a reason to impose a constraint here. What do you think?

p2-apple: It's not something that influences parsing much, at least on our end, so I agree there's not much impetus to be strict here. But I can see that devs choose an arbitrary value, unless they follow an explicit profile, so maybe we can give some guidance, e.g. with something like this?

> The Bundle `type` defaults to  "collection" unless a different type is more appropriate.

jmandel: Yeah. I'll add a link to http://hl7.org/fhir/bundle-definitions.html#Bundle.type and say something like "implementers should choose `collection` unless a more specific type applies". (FHIR rules don't allow for "defaults" here, since it's a required field.)

docs/index.md
jmandel: Done!

p2-apple: Ooooh nice!!

christianpaquin:
Will you bump the version in the changelog, @jmandel ?

PR #240, opened 2023-02-09
Document suggested Deep Link format

p2-apple:
The suggested deep link format uses the URL fragment (delimited with `#`) and starts with an opening parenthesis `(`, followed by the raw JWS, a comma separating any further JWS if necessary, and a closing parenthesis `)`. This allows for percent-encoding free representation of the payload(s) in the fragment and allows detection of the payload type by combining knowledge of the path (suggested as containing `SMARTHealthCard`) and the `()` bounds.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -315,21 +315,21 @@ For a more seamless user experience when FHIR API connections are already in pla
 
 ### via “Deep Link”
 
-So a user can import one or more SMART Health Cards to their Health Wallet with one tap or click, issuers can create app specific “deep links”. These are available on most modern operating systems and will open link clicks in a native app if it's installed on the computer or smartphone.
+For a user to import one or more SMART Health Cards to their Health Wallet with one tap or click, issuers can display app specific “deep links”. These are available on most modern operating systems and will open link clicks in a native app if the respective app is installed on the computer or smartphone.
 
-Even though apps can define their own deep link syntax, for consistency we recommend Health Wallets support the following format:
+Apps can define their own deep link syntax. However, for consistency, we recommend Health Wallets support the following format:
 
 ```
 https://<<associated domain>><<'/'optional path>>/SMARTHealthCard/#(<<Verifiable Credential as JWS>><<','0+ more JWS>>)
 ```
 
-A hypothetical deep link into an app registered for `example.com` with two SMART Health Cards would look like this (actual JWS payload shortened for readability):
+A hypothetical deep link into an app registered for `example.com` with two SMART Health Cards would look something like this (actual JWS payload shortened for readability):
 
 ```text
 https://app.example.com/i/SMARTHealthCard/#(eyJhbGc.dVPBbtswDP.Xo3dhlA,eyJhbGc.xVVNc9MwEP.B3KT7OD)
 ```
 
-The Health Wallet can now parse the one or more JWT and present an import option to the user.
+The Health Wallet app can now parse the JWS and present them for import to the user.
 
 
 #### Discovery of FHIR Support
--- a/docs/index.md
+++ b/docs/index.md
@@ -313,6 +313,25 @@ Finally, the Health Wallet asks the user if they want to save any/all of the sup
 
 For a more seamless user experience when FHIR API connections are already in place, results may also be conveyed through a FHIR API `$health-cards-issue` operation defined [here](../artifacts/operation-patient-i-health-cards-issue.json). For issuers that support SMART on FHIR access, the Health Wallet MAY request authorization with SMART on FHIR scopes (e.g., `launch/patient patient/Immunization.read` for an Immunization use case). This allows the Health Wallet to automatically request issuance of VCs, including requests for periodic updates.
 
+### via “Deep Link”
+
+So a user can import one or more SMART Health Cards to their Health Wallet with one tap or click, issuers can create app specific “deep links”. These are available on most modern operating systems and will open link clicks in a native app if it's installed on the computer or smartphone.
+
+Even though apps can define their own deep link syntax, for consistency we recommend Health Wallets support the following format:
+
+```
+https://<<associated domain>><<'/'optional path>>/SMARTHealthCard/#(<<Verifiable Credential as JWS>><<','0+ more JWS>>)
+```
+
+A hypothetical deep link into an app registered for `example.com` with two SMART Health Cards would look like this (actual JWS payload shortened for readability):
+
+```text
+https://app.example.com/i/SMARTHealthCard/#(eyJhbGc.dVPBbtswDP.Xo3dhlA,eyJhbGc.xVVNc9MwEP.B3KT7OD)
+```
+
+The Health Wallet can now parse the one or more JWT and present an import option to the user.
+
+
 #### Discovery of FHIR Support
 
 A SMART on FHIR Server capable of issuing VCs according to this specification SHALL advertise its support by adding the `health-cards` capability to its `/.well-known/smart-configuration` JSON file. For example:
--- a/docs/index.md
+++ b/docs/index.md
@@ -313,24 +313,27 @@ Finally, the Health Wallet asks the user if they want to save any/all of the sup
 
 For a more seamless user experience when FHIR API connections are already in place, results may also be conveyed through a FHIR API `$health-cards-issue` operation defined [here](../artifacts/operation-patient-i-health-cards-issue.json). For issuers that support SMART on FHIR access, the Health Wallet MAY request authorization with SMART on FHIR scopes (e.g., `launch/patient patient/Immunization.read` for an Immunization use case). This allows the Health Wallet to automatically request issuance of VCs, including requests for periodic updates.
 
-### via “Deep Link”
+### via "Deep Link"
 
-For a user to import one or more SMART Health Cards to their Health Wallet with one tap or click, issuers can display app specific “deep links”. These are available on most modern operating systems and will open link clicks in a native app if the respective app is installed on the computer or smartphone.
+For a user to import one or more SMART Health Cards to their Health Wallet with one tap or click, issuers can display app-specific "deep links". These are available on most modern operating systems and will open in a native app if the respective app is installed on the computer or smartphone.
 
-Apps can define their own deep link syntax. However, for consistency, we recommend Health Wallets support the following format:
+Apps can define their own deep link syntax. However, for consistency we recommend Health Wallets support the following format:
 
 ```
-https://<<associated domain>><<'/'optional path>>/SMARTHealthCard/#(<<Verifiable Credential as JWS>><<','0+ more JWS>>)
+<<app-specific deep link base URL>>#(<<Verifiable Credential as JWS>><<','0+ more JWS>>)
 ```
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -313,6 +313,25 @@ Finally, the Health Wallet asks the user if they want to save any/all of the sup
 
 For a more seamless user experience when FHIR API connections are already in place, results may also be conveyed through a FHIR API `$health-cards-issue` operation defined [here](../artifacts/operation-patient-i-health-cards-issue.json). For issuers that support SMART on FHIR access, the Health Wallet MAY request authorization with SMART on FHIR scopes (e.g., `launch/patient patient/Immunization.read` for an Immunization use case). This allows the Health Wallet to automatically request issuance of VCs, including requests for periodic updates.
 
+### via “Deep Link”
+
+For a user to import one or more SMART Health Cards to their Health Wallet with one tap or click, issuers can display app specific “deep links”. These are available on most modern operating systems and will open link clicks in a native app if the respective app is installed on the computer or smartphone.
```
jmandel, line 318: >https://<<associated domain>><<'/'optional path>>/SMARTHealthCard/#

It's not crystal clear *who* has the option of adding optional paths (I understand it's the server operator but it's implicit here).

Any reason not to instead specify the wallet defines something like a full  "submissions deep link URL, using https or another secure protocol, and ending with `/SMARTHealthCard/#`? Then your syntax can be written relative to the base.

p2-apple: Oh yeah, the app will have to define the path, and your latter paragraph is what I meant – the app will define this and we suggest it ends in `/SMARTHealthCard/#`.

p2-apple: Yes!

jmandel:
OK, I've committed a few changes based on our discussion. Please take a look @p2-apple.

Also just to check: do you have a strong preference for using a payload syntax like:

    #(jws1,jws2)

Instead of

    #{"verifiableCredential": ["jws1","jws2"]}

The latter suggestion would directly reuse our existing JSON file format, and it should be similarly easy to detect + parse. But there may be reasons to prefer the `(`-based syntax. Please let m know.

p2-apple:
Yes, the reason I chose `(` is because that's an allowed character for URL fragments, whereas `{`, `"` and `[` are not. With the suggested `(` syntax we can get by with zero percent encoding, which I think would be neat. Unless we think the overhead of converting the existing file-format syntax to the simplified one is overhead we don't want? I wouldn't think so but happy to hear other opinions.

p2-apple:
Your changes look great to me! If we agree on using `(` then this looks good now.

jmandel:
>  we can get but with zero percent encoding

Hmm. This is true in the sense that somebody creating this deep link would not be *required* to percent in code anything, but the recipient would still need to be capable of handling percent encoded data, right? Because whoever is generating these things could choose to percent in code them, without changing the semantics... Or would you suggest that if we use this custom `(` syntax, we would also prohibit percent encoding from appearing in the fragments?

I guess I think about percent and coding as a fact of life when working with URLs, so it's not particularly costly...

p2-apple:
I don't think we would prohibit that, but with this format the whole link doesn't need any encoding and there's no ambiguity of double encoding/decoding.

p2-apple:
Actual example:

```html
<html>
<head>
    <title>SMART Health Card Link Tester</title>
</head>
<body>
    <div style="max-width: 28em; margin: 3em auto;">
        <h1>Parentheses</h1>
        <p>
            <a href="https://redirect.health.apple.com/SMARTHealthCard/#(eyJhbGciOiJFUzI1NiIsImtpZCI6IjJtMkJJeFhDYXNYTFdndFNyYUg3RlJLMVJoTTQwbXE3UmNxUmN6dzdQY1UiLCJ6aXAiOiJERUYifQ.fZDNTsMwEITfZbkGNymlTX3kXgmJn0uVg-tsGyPHtmwnalX53VknBS7AbWf2085or6BCAA5djC7wxaJDoWPHhHMambQ9FGAOR-DVerMuV9u63rDHh7qsl1UBowR-BemxRROV0C_D4QNlzOaxU_5pMK3GrGjtL8D35A9av3lNgR6DHbxEXlLEl8iwET1O7EmNaGiCnThDk5of7PXiCIFnERWdhvTL7hZeQJy1tFpTN2UNpGKq944-ZMlhxUpWZXtm99_PCL3w8fYRKXwb7mZxnwXd_puTdlRttf2XwTP2bmrUpPQJ.jE8auqQwHLKMpG6eINco0BjVWBbLdWVXoUW9bDKvTaprcLZBCw9GDMPGpDBzH81KHDmsJQBeea61UyOFx2m4Gg,eyJhbGciOiJFUzI1NiIsImtpZCI6IjJtMkJJeFhDYXNYTFdndFNyYUg3RlJLMVJoTTQwbXE3UmNxUmN6dzdQY1UiLCJ6aXAiOiJERUYifQ.fZDNTsMwEITfZbkGNymlTX3kXgmJn0uVg-tsGyPHtmwnalX53VknBS7AbWf2085or6BCAA5djC7wxaJDoWPHhHMambQ9FGAOR-DVerMuV9u63rDHh7qsl1UBowR-BemxRROV0C_D4QNlzOaxU_5pMK3GrGjtL8D35A9av3lNgR6DHbxEXlLEl8iwET1O7EmNaGiCnThDk5of7PXiCIFnERWdhvTL7hZeQJy1tFpTN2UNpGKq944-ZMlhxUpWZXtm99_PCL3w8fYRKXwb7mZxnwXd_puTdlRttf2XwTP2bmrUpPQJ.jE8auqQwHLKMpG6eINco0BjVWBbLdWVXoUW9bDKvTaprcLZBCw9GDMPGpDBzH81KHDmsJQBeea61UyOFx2m4Gg)">Download Vaccinations</a>
        </p>
    </div>
    <div style="max-width: 28em; margin: 3em auto;">
        <h1>JSON Format</h1>
        <p>
            <a href="https://redirect.health.apple.com/SMARTHealthCard/#%7B%22verifiableCredential%22%3A%20%5B%22eyJhbGciOiJFUzI1NiIsImtpZCI6IjJtMkJJeFhDYXNYTFdndFNyYUg3RlJLMVJoTTQwbXE3UmNxUmN6dzdQY1UiLCJ6aXAiOiJERUYifQ.fZDNTsMwEITfZbkGNymlTX3kXgmJn0uVg-tsGyPHtmwnalX53VknBS7AbWf2085or6BCAA5djC7wxaJDoWPHhHMambQ9FGAOR-DVerMuV9u63rDHh7qsl1UBowR-BemxRROV0C_D4QNlzOaxU_5pMK3GrGjtL8D35A9av3lNgR6DHbxEXlLEl8iwET1O7EmNaGiCnThDk5of7PXiCIFnERWdhvTL7hZeQJy1tFpTN2UNpGKq944-ZMlhxUpWZXtm99_PCL3w8fYRKXwb7mZxnwXd_puTdlRttf2XwTP2bmrUpPQJ.jE8auqQwHLKMpG6eINco0BjVWBbLdWVXoUW9bDKvTaprcLZBCw9GDMPGpDBzH81KHDmsJQBeea61UyOFx2m4Gg%22%2C%22eyJhbGciOiJFUzI1NiIsImtpZCI6IjJtMkJJeFhDYXNYTFdndFNyYUg3RlJLMVJoTTQwbXE3UmNxUmN6dzdQY1UiLCJ6aXAiOiJERUYifQ.fZDNTsMwEITfZbkGNymlTX3kXgmJn0uVg-tsGyPHtmwnalX53VknBS7AbWf2085or6BCAA5djC7wxaJDoWPHhHMambQ9FGAOR-DVerMuV9u63rDHh7qsl1UBowR-BemxRROV0C_D4QNlzOaxU_5pMK3GrGjtL8D35A9av3lNgR6DHbxEXlLEl8iwET1O7EmNaGiCnThDk5of7PXiCIFnERWdhvTL7hZeQJy1tFpTN2UNpGKq944-ZMlhxUpWZXtm99_PCL3w8fYRKXwb7mZxnwXd_puTdlRttf2XwTP2bmrUpPQJ.jE8auqQwHLKMpG6eINco0BjVWBbLdWVXoUW9bDKvTaprcLZBCw9GDMPGpDBzH81KHDmsJQBeea61UyOFx2m4Gg%22%5D%7D">Download Vaccinations</a>
        </p>
    </div>
<body>
```

jmandel:
Just to be explicit, the encoding works like this (to take the anchor tag example):

```js
const payload = {verifiableCredential: ["JWS1", "JWS2"]}
const a = document.createElement("a")
a.setAttribute("href", "https://example.com/SMARTHealthCard/#" + encodeURIComponent(payload))
```

Which yields:

    <a href=​"https:​/​/​example.com/​SMARTHealthCard/#%5Bobject%20Object%5D">​</a>​

So yes this involves encoding, but there is no ambiguity. And the recipient needs to decode this, but they'd need to decode the `(` syntax payloads too, just to be safe.

p2-apple:
I'm assuming any URL library will do that anyway when asking for the fragment. Foundation in Swift & ObjC does this for me, at least.

So since either need to be decoded, even if just to be safe, the question remains whether it's more important to stick to the file format or making the link more concise? How many people downloading a SMART Health Card look at the source? 😁

If you think it's worth sticking to the file format then I'm happy to change. On the consuming side it would safe a little bet of decoding work because we can just pipe the fragment into the `.smart-health-card` file decoder.

jmandel:
>How many people downloading a SMART Health Card look at the source?

Definitely nobody. I'm just aiming to "invent" as little as possible, if the existing format can do the job.

That said, I see where you're coming from. Will think on this for a bit.

p2-apple:
The "invent" argument is good. As I said, happy to switch this over. In fact I just did, feel free to continue thinking.

docs/index.md
```diff
@@ -313,6 +313,36 @@ Finally, the Health Wallet asks the user if they want to save any/all of the sup
...
+For a concrete example, consider an app whose deep link base is `https://app.example.com/i/SMARTHealthCard/`. A deep link to import two SMART Health Cards into the app would look something like this (actual JWS payload shortened for readability):
+
+```text
+https://app.example.com/i/SMARTHealthCard/#{"verifiableCredential":["eyJhbGc.dVPBbtswDP.Xo3dhlA","eyJhbGc.xVVNc9MwEP.B3KT7OD"]}
+```
+
+With proper URL encoding a link will look like:
+
+```html
+<a href="https://app.example.com/i/SMARTHealthCard/#%7B%22verifiableCredential%22%3A%20%5B%22eyJhbGc.dVPBbtswDP.Xo3dhlA%22%2C%22eyJhbGc.xVVNc9MwEP.B3KT7OD%22%5D%7D">
```
jmandel, line 339:
```suggestion
<a href="https://app.example.com/i/SMARTHealthCard/#%7B%22verifiableCredential%22%3A%5B%22eyJhbGc.dVPBbtswDP.Xo3dhlA%22%2C%22eyJhbGc.xVVNc9MwEP.B3KT7OD%22%5D%7D">
```

There was an extra `%20` (space) in the encoded example.

jmandel (approved):
With that one tweak (removing the extra `%20` from the example) I think this is good to go.

